PR #16, opened 2024-03-29
Use COO format for edge index

nkakouros:
Currently, the edge indices as provided by `observations['edges']` are a list of tuples with each tuple representing a single edge. However, ML frameworks like pytorch and tensorflow expect the edge indices in COO (coordinate) format. This format is a list of 2 items. Each item is a list containing the starting node's index for each edge. The other list contains the target node. For example:

```
[
    [0, 0, 1, 2]
    [1, 2, 2, 3]
]
```

for a graph with 4 nodes and 4 edges (0-1, 0-2, 1-2, 2-3).

This PR converts the `observations['edges']` list to the coo format.

### DIFF
--- a/malpzsim/sims/mal_petting_zoo_simulator.py
+++ b/malpzsim/sims/mal_petting_zoo_simulator.py
@@ -46,8 +46,8 @@ def _format_full_observation(observation):
             obs_str += header
 
     obs_str += "\nEdges:\n"
-    for edge in observation["edges"]:
-        obs_str += str(edge) + "\n"
+    for index in range(0, len(observation["edges"][0])):
+        obs_str += f"({str(observation["edges"][0][index])}, {str(observation["edges"][1][index])})\n"
 
     return obs_str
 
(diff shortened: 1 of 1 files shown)


kasanari:
Why not just call numpy.transpose? That is what I do in the wrapper to make the edges compatible with pyg. I have nothing against moving it into the obs func, though.

nkakouros:
> Having the list of tuples is IMO a bit easier to iterate over before converting to numpy.

When do you need to iterate over? Also, isn't the edge list in the observations returned by the simulator already a numpy array?

> Why not just call numpy.transpose?

That's what I do, I think it should not be needed though. Eventually, the edge list should be accessible via the AttackGraph itself, it is about the graph anyway. There, the edge list could be stored as a list of tuples and we could also have some helper method/property to read it in coo format, or the other way around. But this here is the simulator that is supposed to be used with ML frameworks; I don't think we should make it more complicated than what is needed to work with it.

kasanari:
> When do you need to iterate over?
It happens once in the changed code. Where instead of "for edge in edges" it iterates over a range of indices. Does not really matter but I find the old version cleaner.

kasanari:
I illustrate how this can be done with np.transpose [here](https://github.com/mal-lang/mal-petting-zoo-simulator/tree/coo_alt)

nkakouros:
> I illustrate how this can be done with np.transpose [here](https://github.com/mal-lang/mal-petting-zoo-simulator/tree/coo_alt)

Nice, that's simpler. Although I still believe the simulator should expose things like edge indices in the "expected" format, this is a low-priority issue and can be better implemented. I am leaving open as a reminder for discussion.

andrewbwm:
This is no longer relevant after the redesign of the MAL Simulator.

PR #17, opened 2024-04-07
Update for changes in maltoolbox

nkakouros:
This PR updates the simulator to follow changes in [this PR](https://github.com/mal-lang/mal-toolbox/pull/13) in the mal-toolbox repo.

A change that I think is particularly useful is the simplified format for the
instance model. It removes a lot of the boilerplate and it is much easier to
write and avoid errors.

### DIFF
--- a/tests/run_demo.py
+++ b/tests/run_demo.py
@@ -2,7 +2,7 @@ from json import JSONEncoder
 import numpy as np
 import logging
 
-from maltoolbox.language import LanguageClassesFactory, LanguageGraph, specification
+from maltoolbox.language import LanguageClassesFactory, LanguageGraph
 from maltoolbox.attackgraph import AttackGraph
 from maltoolbox.model import Model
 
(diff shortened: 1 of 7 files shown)


tests/run_demo.py
```diff
@@ -38,18 +41,15 @@ def default(self, o):
...
 
-model = Model("Test Model", lang_spec, lang_classes_factory)
-model.load_from_file("tests/example_model.json")
+model = Model.load_from_file("tests/example_model.yml", lang_classes_factory)
 
-attack_graph = AttackGraph(lang_spec, model)
-attack_graph.attach_attackers(model)
+attack_graph = AttackGraph(lang_graph, model)
+attack_graph.attach_attackers()
 attack_graph.save_to_file("tmp/attack_graph.json")
```
nkakouros, line 53: I think this should be further simplified:

```
model = Model.load_from_file("tests/example_model.yml", lang_file)
attack_graph = AttackGraph(lang_file, model)
attack_graph.attach_attackers()
```

In other words, the language factory and the language graph should not really be exposed to the users unless they need sth to do with them. It makes the API more consistent too, since the first argument of the model constructor is already a path to a file and not an object.

This is also relevant to this issue: https://github.com/mal-lang/mal-toolbox/issues/21.

PR #34, opened 2024-07-15
Update to new maltoolbox 0.1.2

mrkickling:
This PR adapts MAL Simulator to the newest maltoolbox version.

Note: This also assumes that the maltoolbox bugfixes are merged https://github.com/mal-lang/mal-toolbox/pull/67

What was done:

- Use latest maltoolbox version (0.1.4)
- Refer to type instead of metaconcept
- follow new AttackGraph.load_from_file API
- follow new api for maltoolbox methods where unused args were removed 
- convert the example_model to new format and use it in api_test.py and run_demo
- use get_node_by_full_name instead of by_id
- use fulll_name instead of id when creating the vocab for the Keyboard Agent

### DIFF
--- a/malsim/sims/mal_simulator.py
+++ b/malsim/sims/mal_simulator.py
@@ -585,8 +585,8 @@ class MalSimulator(ParallelEnv):
                     self.agents_dict[agent]["attacker"]
                 ]
                 for node in attacker.reached_attack_steps:
-                    if hasattr(node, "reward"):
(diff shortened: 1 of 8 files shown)


andrewbwm:
Ok, apart from the comment I have already made there are two things that should be done. We should update this once we've released mal-toolbox 0.1.3 and we should use that version for the example_model.yml format because that one does properly only include non-default defenses.

PR #37, opened 2024-07-25
Testing

mrkickling:
- Tests for MalSimulator methods (pytest)


mrkickling:
This would be nice to merge so we can see that the malsimulators behavior is consistent when we make changes to it.

PR #50, opened 2024-09-11
Scenario in gym wrapper

mrkickling:
Make use of scenario in GymWrappers, replaces the LazyWrapper.


malsim/scenario.py
```diff
@@ -32,13 +33,13 @@
     'lang_file',
     'model_file',
     'attacker_agent_class',
-    'defender_agent_class',
```
andrewbwm, line 35: We've discussed this previously. I would always have both types of agents be required and you can just set them to `None`/`False` if you don't want one.

This is because I don't want people to accidentally not provide certain agents in their scenarios. They should explicitly state that that scenario does not contain an Attacker or Defender.

malsim/scenario.py
```diff
@@ -106,6 +114,41 @@ def apply_scenario_attacker_entrypoints(
 
         attacker.entry_points = attacker.reached_attack_steps
 
+def load_scenario_simulation_config(scenario: dict):
```
andrewbwm, line 117: Add the typehint for the return too.

malsim/sims/mal_simulator.py
```diff
@@ -293,6 +293,7 @@ def action_space(self, agent=None):
         return MultiDiscrete([num_actions, num_objects], dtype=np.int64)
 
     def reset(self, seed: Optional[int] = None, options: Optional[dict] = None):
+        # TODO: params not used by method
```
andrewbwm, line 296: I think the reset signature is defined by the PettingZoo or Gymnasium APIs, and we're just making it be compatible.

malsim/sims/mal_simulator.py
```diff
@@ -393,6 +394,16 @@ def register_defender(self, agent_name):
         self.possible_agents.insert(0, agent_name)
         self.agents_dict[agent_name] = {"type": "defender"}
 
+    def get_attacker_agents(self):
+        """Return agents_dict but only with attacker agents"""
+        return {k: v for k, v in self.agents_dict.items()
+                if v['type'] == "attacker"}
+
+    def get_defender_agents(self):
+        """Return agents_dict but only with defender agents"""
```
andrewbwm, line 403: Rephrase comment to Return agents dictionaries of attacker agents.

And the one bellow Return agents dictionaries of defender agents.

andrewbwm (changes requested):
Apart from the minor things that I flagged this looks great!

malsim/scenario.py
mrkickling: I think it makes sense, will revert it.

malsim/sims/mal_simulator.py
mrkickling: That makes sense for the signature, but we probably don't need to send them to MalSimulator.reset() in AttackerEnv.reset() and DefenderEnv.reset then. Not sure though.

malsim/scenario.py
mrkickling: Resolved

malsim/sims/mal_simulator.py
mrkickling: Added TODO to Defender/AttackerEnv.reset instead

malsim/sims/mal_simulator.py
mrkickling: Fixed

kasanari:
Looks good.

kasanari:
```
    malsim.register_envs()
    scenario_file = "/storage/GitHub/malsim-scenarios/scenarios/trainingLang/scenario-phishing-one-host/scenario.yml"
    env = gym.make("MALDefenderEnv-v0",
        scenario_file=scenario_file,
    )

    attack_graph, _ = load_scenario(scenario_file)
```

Found a potential bug. When I load the scenario with the wrapper, the attack graph has six nodes, but then I load the same scenario with the load_scenario function it has seven.

andrewbwm:
> ```python
> malsim.register_envs()
> scenario_file = "/storage/GitHub/malsim-scenarios/scenarios/trainingLang/scenario-phishing-one-host/scenario.yml"
> env = gym.make("MALDefenderEnv-v0",
>     scenario_file=scenario_file,
> )
> 
> attack_graph, _ = load_scenario(scenario_file)
> ```
> 
> Found a potential bug. When I load the scenario with the wrapper, the attack graph has six nodes, but then I load the same scenario with the load_scenario function it has seven.

That is because the MalSimulator initialization prunes the unnecessary and unviable nodes right now. The load_scenario method does not create the simulator just yet, which is why they are of different sizes.

mrkickling:
Feels like a design desition, should the graph be pruned by default or not? See #40

andrewbwm:
> Feels like a design decision, should the graph be pruned by default or not? See #40

I think it should be enabled by default because most people don't like seeing untraversable nodes and they might not dig enough to figure out that pruning is even an option.

PR #57, opened 2024-09-23
Simulator settings

mrkickling:
Implement MalSimulatorSettings

A dataclass that can be created and given to `MalSimulator.__init__` when creating a MalSimulator. Contains default values, so if no MalSimulatorSettings object is given, things will still work as expected (see defaults below)

Currently supports two configuration options:

```python
    # uncompromise_untraversable_steps
    # - Uncompromise (evict attacker) from nodes/steps that are no longer
    #   traversable (often because a defense kicked in) if set to True
    # otherwise:
    # - Leave the node/step compromised even after it becomes untraversable
    uncompromise_untraversable_steps: bool = False

    # cumulative_actions_in_defender_obs
    # - Defender sees the status of the whole attack graph if set to True
    # otherwise:
    # - Defender only sees the status of nodes changed in the current step
    cumulative_defender_obs: bool = True
```

I need help with the naming (Andrei already helped a bit)
and whether it makes sense to have it as a class or not.

Also please check if the implementation of `cumulative_actions_in_defender_obs` looks sane.
I needed a way to hand the latest performed actions to the _observe_defender for it to work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ through the parameter 'sim_settings'. Giving sim_settings is optional, otherwise
 
 settings = MalSimulatorSettings(
   uncompromise_untraversable_steps=True, # default is False
-  cumulative_actions_in_defender_obs=False # default is True
+  cumulative_defender_obs=False # default is True
 )
 sim = MalSimulator(lang_graph, model, attack_graph, sim_settings=settings)
 
--- a/README.md
+++ b/README.md
@@ -7,7 +7,23 @@ A MAL compliant simulator.
 ## Installation
 ```pip install mal-simulator```
 
-Use the malsim CLI to run run simulations on scenarios.
+## MalSimulator
+
+A `sims.mal_simulator.MalSimulator` can be created to be able to run simulations.
+
+### MalSimulatorSettings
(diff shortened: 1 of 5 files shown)


kasanari:
I think the names are good. `cumulative_actions_in_defender_obs` might be shorted to `cumulative_defender_obs`.

One suggestion: If we are adding partial observability to the defender (this includes false positives), is to add a third "agent" to the obs that contains the true state or at least have a function that makes it easy to get it. Cage 2 has this, for example: https://github.com/cage-challenge/cage-challenge-2/blob/26ce1c1253fa9e2e73f25e6a7f2da32860c11257/CybORG/CybORG/CybORG.py#L123

mrkickling:
@kasanari good suggestion on the name. 
By true state you mean the full state of the attack graph in this case?
What do you mean by a third 'agent' ? A method for it I understand but 'agent' is unclear to me in this context.

kasanari:
I would be a third key in the observation dict. Currently it is "attacker" and "defender". The true state could just be "state".

mrkickling:
Then I get it. A predefined agent key in the observation which contains observed_state which is the true state of the attack graph (with all information). Maybe it could even be called 'true_state'. I will make an issue for it.

PR #61, opened 2024-09-24
Factor out debug tables

mrkickling:
- Factor out into separate method
- Make sure it is only calculated on debug logging level

### DIFF
--- a/malsim/sims/mal_simulator.py
+++ b/malsim/sims/mal_simulator.py
@@ -310,17 +310,12 @@ class MalSimulator(ParallelEnv):
         self.attack_graph = copy.deepcopy(self.attack_graph_backup)
         return self.init(self.max_iter)
 
-    def init(self, max_iter=ITERATIONS_LIMIT):
-        logger.info("Initializing MAL ParralelEnv Simulator.")
-        logger.debug("Creating and listing mapping tables.")
-        self._index_to_id = [n.id for n in self.attack_graph.nodes]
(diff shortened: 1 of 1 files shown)


andrewbwm:
Looks good, I suspect you're already working on this, but I'll add it as a reminder just in case. We should also rework the string formatting so that it uses c-type printing instead of the format() str method.

mrkickling:
Yes, and I actually postponed it at this time (#62) because it feels too messy to do it while we refactor other stuff in malsim.

andrewbwm:
> Yes, and I actually postponed it at this time (#62) because it feels too messy to do it while we refactor other stuff in malsim.

Yes, I saw that, and that makes sense, but you should have the ones specifically used with format_table match the preferred type.

mrkickling:
I looked into it, but since headers are strings and not the values, it becomes problematic. Also "%-5d %-80s %-6d %-5d %-5d %-5s %-5d %s\n" is not more readable than the alternative .. maybe for the table creation using the .format is not actually that bad?

andrewbwm:
> I looked into it, but since headers are strings and not the values, it becomes problematic. Also "%-5d %-80s %-6d %-5d %-5d %-5s %-5d %s\n" is not more readable than the alternative .. maybe for the table creation using the .format is not actually that bad?

Ok, we can leave it as a separate issue an look into it when we fix the entire logging.

mrkickling:
Then I will merge if that is alright for you

andrewbwm (approved):

PR #67, opened 2024-10-03
False positives negatives rate

mrkickling:
User can specify false positive/negative rate for specific attack graph nodes
The are applied by scenario on the attack graph into node.extras field.

They do not affect simulation run, but are supposed to be used by wrappers that build observations.

NOTE: ignore previous comments here because that was regarding the old serialized environment

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,24 @@ actionable_steps:
   #       - phishing
   #     ...
 
+
+# Optionally add false positive/negative rates to observations.
+#
+# False positive/negative `rate` is a number between 0.0 and 1.0.
+#  - A false positive rate of x for means that an inactive attack step
+#    will be observed as active at a rate of x in each observation
+#  - A false negative rate of x for means that an active attack step
+#    will be observed as inactive at a rate of x in each observation
+# Default false positive/negative rate is 0, which is assumed if none are given.
+
+# Applies false positive rates per attack step (default 0)
+false_positive_rates:
+  <attack step full name>: <rate>
+
+# Applies false negative rates per attack step (default 0)
+false_negative_rates:
+  <attack step full name>: <rate>
+
 ```
 
 ### Loading a scenario from a python script
--- a/README.md
+++ b/README.md
@@ -125,11 +125,21 @@ actionable_steps:
 
 # Applies false positive rates per attack step (default 0)
 false_positive_rates:
-  <attack step full name>: <rate>
+  by_asset_type:
+    <asset_type>:
+      <step name>: rate (float)
+  by_asset_name:
+    <asset_name>:
+      <step name>: rate (float)
 
 # Applies false negative rates per attack step (default 0)
 false_negative_rates:
-  <attack step full name>: <rate>
+  by_asset_type:
+    <asset_type>:
+      <step name>: rate (float)
+  by_asset_name:
+    <asset_name>:
+      <step name>: rate (float)
 
 ```
 
--- a/malsim/scenario.py
+++ b/malsim/scenario.py
@@ -108,7 +108,7 @@ def apply_scenario_rewards(
         node.extras['reward'] = reward
 
 
-def _validate_scenario_property_rules(
+def _validate_scenario_node_property_config(
         graph: AttackGraph, rules: dict):
     """Verify that observability/actionability rules in a scenario contains
     only valid assets, asset types and step names"""
@@ -179,7 +179,7 @@ def apply_scenario_node_property_rules(
              and/or `by_asset_type`
     """
 
-    _validate_scenario_property_rules(attack_graph, rules)
+    _validate_scenario_node_property_config(attack_graph, rules)
 
     if not rules:
         # If no rules are given, make all steps as observable/actionable
@@ -210,7 +210,7 @@ def apply_scenario_node_property_rules(
 def apply_scenario_node_property_values(
         attack_graph: AttackGraph, node_prop: str, values: dict,
 ):
-    """Apply node property rules from scenario configuration.
+    """Apply node property values from scenario configuration.
 
     Sets node.extras[node_prop] to value defined by rules for nodes
     that match rule in the AttackGraph. Values defined per asset has
@@ -223,7 +223,7 @@ def apply_scenario_node_property_values(
               and/or `by_asset_type`
     """
 
-    _validate_scenario_property_rules(attack_graph, values)
+    _validate_scenario_node_property_config(attack_graph, values)
 
     if not values:
         # If no rules defined, do not set the values at all
--- a/malsim/scenario.py
+++ b/malsim/scenario.py
@@ -56,8 +56,6 @@ allowed_fields = required_fields + [
     'rewards',
     'observable_steps',
     'actionable_steps',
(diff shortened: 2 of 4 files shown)


malsim/sims/mal_simulator.py
```diff
@@ -654,9 +655,15 @@ def _observe_defender(
                     observation["observed_state"][index] = 1
                 else:
                     if node.is_compromised():
-                        observation["observed_state"][index] = 1
+                        fn_rate = node.extras.get('false_negative_rate')
+                        make_false_negative = fn_rate and random() < fn_rate
```
kasanari, line 659: I would advise using a random number generator object here, so that the simulator respects the given seed.

kasanari (commented):
Looks fine apart from RNG management.

malsim/sims/mal_simulator.py
mrkickling: Make sense, will fix that.

mrkickling: I made a lot of changes to the PR after restructuring was done in the main branch. Needs to be reviewed again.

mrkickling: I also added a test for the rng functionality to make sure a seed makes it deterministic.

tests/test_mal_simulator.py
```diff
@@ -956,6 +957,140 @@ def test_simulator_default_settings_defender_observation():
...
+            seed,
+            expected_attacker_obs,
+            expected_defender_obs
+        ):
+        sim.reset(seed=seed)
+        attacker_agent_id = next(iter(sim.get_attacker_agents()))
+        defender_agent_id = next(iter(sim.get_defender_agents()))
+
+        # Get an uncompromised step
+        user_3_compromise = sim.attack_graph.get_node_by_full_name(
```
kasanari, line 1007: This is good to check, but it would be even better if an actual attacker agent is used, or action_space.sample

kasanari (approved):

tests/test_mal_simulator.py
mrkickling: Do you mean to test the attacker agent seed as well? I think we should have a test for that separately.

kasanari: Perhaps both, even? One for the attacker and one for the integration of both the simulator and attacker.

mrkickling: I added tests for the attacker agents now

mrkickling: Added test for the combination of giving seed to agent AND simulator AND having fp/fn rates.

README.md
```diff
@@ -109,6 +109,24 @@ actionable_steps:
   #       - phishing
   #     ...
 
+
+# Optionally add false positive/negative rates to observations.
+#
+# False positive/negative `rate` is a number between 0.0 and 1.0.
```
kasanari, line 115: False positive implies that it is actually negative, so it is probability of positive given that it is negative. Inverse for false negative rate.

tests/testdata/scenarios/traininglang_fp_fn_scenario.yml
```diff
@@ -0,0 +1,34 @@
...
+
+# Specific fpr/fnr for some specific steps
+false_positive_rates:
+  'Host:0:access': 0.2
+  'Host:1:access': 0.3
+
+false_negative_rates:
+  'Host:0:access': 0.4
+  'Host:1:access': 0.5
+  'User:3:compromise': 1.0 # Always false negative
```
kasanari, line 22: You could add a test that runs through the attack graph with an attacker and checks the expected occurrence of the steps. For instance, the false negative = 1 is always invisible. FPR = 1 and FNR = 1 leads to the step being fully observable, but in inverse where it always shows as negative until the attacker compromises it upon which it becomes zero.

kasanari: You could also test a fixed series of attack steps and confirm that an attack step outside of that sequence shows up due to FPR, or that some of them do not due to FPR.

malsim/scenario.py
```diff
@@ -55,7 +55,11 @@
 allowed_fields = required_fields + [
     'rewards',
     'observable_steps',
-    'actionable_steps'
+    'actionable_steps',
+    'attacker_entry_points',
+    'observable_attack_steps',
+    'false_positive_rates',
+    'false_negative_rates',
 ]
```
nkakouros, line 63: The `*_fields` variables need updating. For example, you list `attacker_entry_points` in allowed_fields while it is in `deprecated_fields` (makes sense), but you omit the other 2 deprecated fields. Also, `observable_steps` seems to be the same with `observable_attack_steps`? But neither of them is used anywhere in the code. 

On a side note, it would be more robust to use jsonschema (or other) to do the scenario file validation. It would also act as documentation for how to write scenario files.

malsim/scenario.py
```diff
@@ -204,6 +208,34 @@ def apply_scenario_node_property_rules(
...
+    fn_rates_per_attackstep = scenario_conf.get('false_negative_rates')
+    if fn_rates_per_attackstep:
+        for step_full_name, rate in fn_rates_per_attackstep.items():
+            step = attack_graph.get_node_by_full_name(step_full_name)
+            assert step, (
+                f"Attack step {step_full_name} not found in attack graph"
+            )
+            step.extras['false_negative_rate'] = rate
+
+
```
nkakouros, line 238: The implementation is quite generic in the sense that if you replace the words about fp_rates/fn_rates with sth else that assigns a value to a step, the code will still work. Probably, it does not make sense to generalize this now, but for the future it could be sth to consider.

mrkickling: Yes, it might be worth doing it now, and making it more similar to how the actionability/observability is set.

malsim/scenario.py
```diff
@@ -193,17 +192,72 @@ def apply_scenario_node_property_rules(
...
             if step.name in node_prop_rule_step_names:
                 step.extras[node_prop] = 1
             else:
                 step.extras[node_prop] = 0
 
 
+def apply_scenario_node_property_values(
+        attack_graph: AttackGraph, node_prop: str, values: dict,
+):
+    """Apply node property rules from scenario configuration.
```
mrkickling, line 215: property rules -> property values

malsim/scenario.py
```diff
@@ -193,17 +192,72 @@ def apply_scenario_node_property_rules(
...
+    precedence over nodes defined per asset type.
+
+    Arguments:
+    - attack_graph: The attack graph to apply the settings to
+    - node_prop: property name in string format (i.e. 'false_positive_rate')
+    - values: settings from scenario file with keys `by_asset_name`
+              and/or `by_asset_type`
+    """
+
+    _validate_scenario_property_rules(attack_graph, values)
```
andrewbwm, line 228: `_validate_scenario_property_values`

malsim/scenario.py
mrkickling: I will have the deprecated fields only be in `deprecated_fields` and remove the unused `observable_attack_steps`

andrewbwm (approved):

PR #68, opened 2024-10-07
Restructure observed states in simulator

mrkickling:
For performance and structural reasons we want to change how states are built in MalSimulator.
Currently, observations (observed states) are built up every time. This is costly, especially for the attacker as they enable more and more attack steps which leads to us looping over more and more nodes in each step of the simulator.

The proposed solution is to instead build up the state in the beginning of the simulation and then add on the delta (actions) performed in each step to each agents observed state depending on their type., This definitely gives a performance boost, and will improve training time as well (how much?).

I have marked many questions with TODO in the code that need to be sorted out before merge.

We also need more tests and review to make sure no behavior has changed in the Simulator due to this PR.

### DIFF
--- a/malsim/sims/mal_simulator.py
+++ b/malsim/sims/mal_simulator.py
@@ -563,27 +563,28 @@ class MalSimulator(ParallelEnv):
             for agent in self.agents:
                 if self.agents_dict[agent]["type"] == "attacker":
 
-                    attacker_index = self.agents_dict[agent]["attacker"]
-                    attacker = self.attack_graph.attackers[attacker_index]
-                    node_index = self._id_to_index[node.id]
+                    if self.sim_settings.uncompromise_untraversable_steps:
+                        attacker_index = self.agents_dict[agent]["attacker"]
+                        attacker = self.attack_graph.attackers[attacker_index]
+                        node_index = self._id_to_index[node.id]
 
-                    logger.info(
-                        'Remove untraversable node from attacker '
-                        '"%s": "%s"(%d)',
-                            agent,
-                            node.full_name,
-                            node.id
-                    )
+                        logger.info(
+                            'Remove untraversable node from attacker '
+                            '"%s": "%s"(%d)',
+                                agent,
+                                node.full_name,
+                                node.id
+                        )
 
-                    agent_observation = self._agent_observations[agent]
-                    node_obs = agent_observation['observed_state'][node_index]
+                        agent_observation = self._agent_observations[agent]
+                        node_obs = agent_observation['observed_state'][node_index]
 
-                    if node.type in ('and', 'or') and \
-                       node_obs == 1 and \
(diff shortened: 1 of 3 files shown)


malsim/sims/mal_simulator.py
```diff
@@ -566,126 +649,124 @@ def _defender_step(self, agent, defense_step):
...
+                    # If attacker performed step, it will receive
+                    # a reward and penalize all defender
+                    self._agent_rewards[agent] += node_reward
+
+                    for d_agent in self.get_defender_agents():
+                        self._agent_rewards[d_agent] -= node_reward
+                else:
+                    # TODO: defending against an attack step
+                    # can it increase reward for defender?
+
```
andrewbwm, line 749: Well, technically it increases the penalty, since it's negative. Apart from that, yes, that's how defences work. They cost something, so the defender is penalized if there is a reward associated with it. Which makes sense, if there's any defence without a cost then the correct action is just to always enable it immediately.

malsim/sims/mal_simulator.py
```diff
@@ -420,15 +469,21 @@ def register_attacker(self, agent_name, attacker: int):
...
-            "attacker": attacker}
+        self.agents_dict[agent_name] = {
+            "type": "attacker",
+            "attacker": attacker
+        }
 
     def register_defender(self, agent_name):
-        # Defenders are run first so that the defenses prevent the attacker
-        # appropriately in case the attacker selects an attack step that the
-        # defender safeguards against in the same step.
```
andrewbwm, line 430: Why was this removed?

malsim/sims/mal_simulator.py
```diff
@@ -795,7 +878,11 @@ def step(self, actions):
             "Stepping through iteration %d/%d", self.cur_iter, self.max_iter)
         logger.debug("Performing actions: %s", actions)
 
-        performed_actions = []
+        # Map agent to defense/attack steps performed in this step
+        performed_actions = {}
+
+        # TODO: where do we make sure defender is run before attacker?
+
```
andrewbwm, line 885: Line 487 of the new code in `register_defender`. It used to have a docstring explaining this but you removed it ... irony. :)

malsim/sims/mal_simulator.py
```diff
@@ -373,8 +374,66 @@ def _create_mapping_tables(self):
             n: i for i, n in enumerate(self._index_to_step_name)
         }
 
+    def _initialize_agents(self) -> dict[str, list[int]]:
+        """Initialize agent rewards, observations and action_surfaces
+
+        Updates self._agent_rewards, self._agent_observations
+        and self.action_surfaces
```
andrewbwm, line 381: This feels redundant and kind of implementation specific in a non-useful way.

andrewbwm (approved):

PR #73, opened 2024-10-21
Add heuristics defender agent

mrkickling:
- Add Heuristics agent which does whack-a-mole
- Make agents a package
- Add base class for MalSimulatorAgents
- Test scenario with heuristics agent


andrewbwm:
This was implemented in the tyr-monitor instead.

PR #80, opened 2024-11-28
Add base agent class

mrkickling:


malsim/agents/agent_base.py
```diff
@@ -0,0 +1,22 @@
...
+        )
+
+    @abstractmethod
+    def compute_action_from_dict(
+        self, observation: dict, mask: dict) -> tuple[int, Optional[int]]:
+        """From observation and mask, find the next action for agent"""
+
+        raise NotImplementedError(
+            "compute_action_from_dict must be implemented by inheriting class"
+        )
```
andrewbwm, line 22: Only machine learning agents should have this, not base agents.

andrewbwm: Sort of, the name isn't the biggest issue, the parameters that come with the signature are. I disagree with the approach, I think you should break them, we don't want that compability with the old agents since they're not really implemented at the correct level. Doing it partially like this just feels confusing.

mrkickling: Yes, it should instead take the action_surface (list of nodes) I believe.

Ok, this is work that I have already been doing on a different branch.
Do you think I should just have this PR (with correct signatures) be part of that branch and have it all merged at once at a later stage?

andrewbwm: Yes, I think they should be part of the same larger restructuring PR. And I also don't think it should be given the action surface as a parameter. I think these agents should be initialized with an attacker and then they just internally use the action surface of that attacker. Does that make sense?

mrkickling: BFS/DFS could be both attacker and defender for example. HeuristicsDefender makes no sense to use as an attacker on the other hand.

andrewbwm: Right, that makes some sense. But I do think what we want is to actually separate them based on [Attacker/Defender/Observer]Agents. In which case they each have their surfaces they look into which wouldn't be passed as parameters. It's not a trivial question, because passing it as a parameter makes it easier to change it just for one agent instead of having to change the underlying attack graph, but it also makes it more error prone.

PR #87, opened 2024-12-17
Simplify MalSimulator and create wrappers for other interfaces

mrkickling:
This is a rather big redesign of the simulator.

MalSimulator is now very much simplified:
- This creates a base for other wrappers (with their own observation formats) to use
- It also allows to run the simulator more easily for other reasons than ML training
- It works on AttackGraphs only
- It stores agent states (Agent classes were added for this)
- The MAL Simulator contains the state of the Attack Graph
- step() takes AttackGraphNodes as actions and returns those that were performed (not the full state). Does this make sense?

Things related to building up observations (specifically the ParallelEnv part that before was in the MalSimulator) has now been moved to a wrapper `malsim_vectorized_obs_env.py` (naming here is an open question).
This wrapper builds up state step by step from the performed actions returned from MalSimulator.step(actions).
Logging for that wrapper was also factored into its own module (should it be?)

Added the class DecisionAgent (these are agents like BFS/DFS + keyboard + in the future more advanced heuristics).
This was to create a common interface for working with DecisionAgents.
All our DecisionAgents are still not done, they are still tailored towards the ParallelEnv/MalSim Vectorized Obs Env, but the goal is to have them work with the regular MalSimulator (on attack graphs state).

Any number of agents can now be specified in the scenario file. This is a cleaner solution in my opinion.

Let me know if something does not make sense.


malsim/sims/mal_simulator.py
```diff
@@ -1,70 +1,78 @@
...
 
-    formatted_str = ''
-    header = entry_format.format(*header_entry)
-    formatted_str += header
-    for entry_nr, entry in zip(range(0, len(entries)), entries):
-        formatted_str += entry_format.format(*entry)
-        if (reprint_header != 0) and ((entry_nr + 1) % reprint_header == 0):
-            formatted_str += header
-    return formatted_str
+@dataclass
```
kasanari, line 27: Having this be a non-mutable NamedTuple would be safer IMO

nkakouros: The dataclass can also become immutable with `@dataclass(frozen=True)`.

mrkickling: Sorry, why do we want them to be immutable?
The agent states are (at least in the current implementation) stored in the simulator and potentially updated/changed each step.

kasanari: Are they not also sent outside the simulator, as per the function signature of the searchers? Sending references to mutable state outside of the simulator seems dangerous to me.

mrkickling: I see your point now. As you say the agent states are not only internal, but used by the decision agents/searchers. I agree that it feels dangerous. Maybe they should be immutable, but it also feels a bit weird to create new Agent objects in every step. Other option is to use something else in the searchers than those objects, and keep them internal.

nkakouros: You can do:

```python
@property
def state(self):
    """Immutable state"""
    return tuple(self._mutable_state)
```

or sth similar.

But the current implementation is not safe in this manner in a thousand places. It is also very easy for someone to bypass these safety measures. If it is too cumbersome to create such a dynamic property (because converting the internal data structure - a dataclass in this case - to sth immutable is more involved), then it is not worth it. In this case, the simplest approach seems to be to [convert the dataclass to a frozen one](https://python-forum.io/thread-43340.html) when reading the property. But I don't know the performance implications.

mrkickling: Yes, having the agent state object be mutable internally in the MalSimulator but return an immutable object when fetching it might be the easiest way. It would avoid the dangers and confusion of the agent states being changed outside of the simulator. Does that make sense?

kasanari: That seems like an appropriate solution.

mrkickling: I added MalSimAgentView (a public interface for MalSimAgent). So every time an agent state is returned from MalSimulator it will be read-only.

The observations/infos in the SerializedObsEnv are now created without altering the MalSimAgent objects and instead stored in dicts.

Bonus explanation about the searcher agents:
The searchers still depend on action_mask (which comes from the SerializedObsEnv) which is not what I/we actually want in the end (we want the decision agents to work with the MalSimulator without serialization/indexing/action masks). The problem was that when I converted the searcher agents to use the agent state only they behaved a little bit different from before, so before I solve that I will let them stay that way.

malsim/agents/searchers.py
```diff
@@ -81,7 +85,7 @@ def select_next_target(
         return current_target, False
 
 
-class DepthFirstAttacker:
+class DepthFirstAttacker():
```
Hoclor, line 88: Shouldn't the DepthFirstAttacker also inherit DecisionAgent?

malsim/agents/decision_agent.py
nkakouros, line 1: How about we consolidating the design of the agents taking into account my PR [here](https://github.com/mal-lang/mal-simulator/pull/90) too? Otherwise, if you plan to merge that too, it does not make sense to have them redesigned here only to be redesigned again later.

mrkickling: I had not seen your PR. 

In my opinion the agents should not depend on the obs/action mask. Partly because it is not needed and partly because the obs/action mask will be specific to the MalSimSerializedEnv/ParalellEnv and not a part of the base MalSimulator (as per this PR).

The searcher agents in this PR only depend on the action surface of an agent (a list of AttackGraphNodes). I am open to other ways and of simplifying the searcher agents further, but I do want to get rid of using the obs/action mask. What do you think?

nkakouros: I'm all for removing these cryptic stuff. My comment was mostly about the design of the agents which seems unnecessarily convoluted in its current form. There is really no need for 4 5 methods just to pass things around or 5 variables to do what can be done with 2, etc. 

Specifically to your changes here, is there an ambition behind the Decision agent? To me, as long as the interface of an agent is fixed and explicit (i.e. the init and the get-next-step-what-its-called-method) there is really no benefit in providing abstract classes or trying to standardize the internal structure of an agent class, especially when it is that little. Or are there requests for sth like that by others that make such things a necessity or sth? 

My approach would be to take my PR (because I think it's great 😃) and just remove the mask etc, unless I'm missing sth bigger.

nkakouros: Also, I think it doesn't offer much to have three files to keep what is basically 200 lines of code.  The perceived benefit of having the agent module be a package where things are split using a conceptual separation does not add much in practice, it makes importing a tiny bit slower, requires 3 open files if someone is editing the package in an editor, etc, and more importantly is undone by exposing classes at the package level using `__all__` in `__init__`.

mrkickling: Yeah, I am all for simplifying it into one or two method if it is possible. I also added two simple tests so that makes it safer to make those changes.

If there is an ambition behind using the abstract base class DecisionAgent?
The main reason is that it makes it clear for the user/developer that every class that implements it shares the same interface and has the same method name. I agree that it might be a bit overkill since there will probably just be one method.. but at the same time I don't think it hurts, especially when using a linter it gets clear that for example in cli.py we can use `decision_agent.get_next_action(sim_agent_state)` even though we don't know what class decision_agent has (we type hint it with `DecisionAgent`). 

Maybe it is not super pythonic, and maybe there is some better option? Or do you think it adds no value?

mrkickling: I didn't see the second comment before sending that.

I think having separated files  makes sense in this case for future proofing instead of putting all agents in one class. We will probably add heuristics agents in there (agents that take decision based on the state of the attack graph) that could be in a separate file and we might add searcher agents which are specific for defenders as well, so it will grow. 

But yeah, if it makes sense to have a base DecisionAgent is more questionable.

kasanari:
Throwing this into the mix:

```python
# Author: Jakob Nyberg, 2025
from collections.abc import Callable
import logging


from typing import Any

import numpy as np
import numpy.typing as npt

logger = logging.getLogger(__name__)


Array = npt.NDArray[np.int32]
BoolArray = npt.NDArray[np.bool_]


def get_new_targets(
    discovered_targets: npt.NDArray[np.int32],
    mask: tuple[npt.NDArray[np.bool_], npt.NDArray[np.bool_]],
) -> tuple[npt.NDArray[np.int32], npt.NDArray[np.int32]]:
    attack_surface = mask[1]
    surface_indexes = np.flatnonzero(attack_surface)
    new_targets = np.array(
        [idx for idx in surface_indexes if idx not in discovered_targets],
        dtype=np.int32,
    )
    return new_targets, surface_indexes


def move_target_to_back(
    current_target: np.int32 | None,
    targets: npt.NDArray[np.int32],
    attack_surface: npt.NDArray[np.int32],
) -> tuple[npt.NDArray[np.int32], np.int32 | None]:
    """
    If the current target was not compromised this turn, put it
    on the bottom of the stack and focus on next target instead
    """

    if not current_target:
        return targets, current_target

    if current_target in attack_surface:
        targets = np.concatenate((current_target, targets[:-1]))
        return targets, targets[-1]

    return targets, current_target


def choose_target(
    targets: npt.NDArray[np.int32],
    attack_surface: npt.NDArray[np.int32],
) -> tuple[npt.NDArray[np.int32], np.int32, bool]:
    # targets that have not been removed from the attack surface by another agent
    valid_targets = np.array(
        [t for t in targets if t in attack_surface], dtype=np.int32
    )

    if len(valid_targets) == 0:
        return valid_targets, np.int32(0), True

    return valid_targets[:-1], valid_targets[-1], False


def compute_action(
    permute_func: Callable[[Array], Array],
    action_func: Callable[
        [np.int32 | None, Array, Array], tuple[Array, np.int32 | None, bool]
    ],
    add_targets_func: Callable[[Array, Array], Array],
):
    def _compute_action(
        targets: Array,
        mask: tuple[BoolArray, BoolArray],
        current_target: np.int32 | None,
    ):
        new_targets, surface_indexes = get_new_targets(targets, mask)

        targets, current_target, done = action_func(
            current_target,
            add_targets_func(targets, permute_func(new_targets)),
            surface_indexes,
        )

        action = 0 if done else 1
        if action == 0:
            logger.debug(
                'Attacker agent does not have any valid targets and will terminate'
            )

        logger.debug(f'Attacker targets: {targets}')
        logger.debug(f'Attacker current target: {current_target}')
        logger.debug(f'Attacker action: {action}')
        return targets, action, current_target

    return _compute_action


def create_permute_func(seed: int | None, randomize: bool) -> Callable[[Array], Array]:
    s = seed if seed else np.random.SeedSequence().entropy
    rng = np.random.default_rng(s) if randomize else None
    return rng.permutation if rng else lambda x: x


class BreadthFirstAttacker:
    def __init__(self, agent_config: dict[str, Any]) -> None:
        self.current_target: np.int32 | None = None
        self.targets: npt.NDArray[np.int32] = np.array([], dtype=np.int32)

        permute_func = create_permute_func(
            agent_config.get('seed', None), agent_config.get('randomize', False)
        )

        self.compute_action = compute_action(
            permute_func, self._action_func, self._add_new_targets_func
        )

    def __call__(
        self,
        _: dict[str, Any],
        mask: tuple[npt.NDArray[np.bool_], npt.NDArray[np.bool_]],
    ):
        self.targets, action, self.current_target = self.compute_action(
            self.targets, mask, self.current_target
        )

        return (action, self.current_target)

    @staticmethod
    def _action_func(
        current_target: np.int32 | None,
        targets: Array,
        surface_indexes: Array,
    ):
        targets, current_target = move_target_to_back(
            current_target, targets, surface_indexes
        )
        return choose_target(targets, surface_indexes)

    @staticmethod
    def _add_new_targets_func(targets: Array, new_targets: Array):
        new_targets = np.flip(new_targets)  # to comply with the original implementation
        return np.concatenate([new_targets, targets])


class DepthFirstAttacker:
    def __init__(self, agent_config: dict[str, Any]) -> None:
        self.current_target: np.int32 | None = None
        self.targets: npt.NDArray[np.int32] = np.array([], dtype=np.int32)

        permute_func = create_permute_func(
            agent_config.get('seed', None), agent_config.get('randomize', False)
        )

        self.compute_action = compute_action(
            permute_func, self._action_func, self._add_new_targets_func
        )

    def __call__(
        self,
        _: dict[str, Any],
        mask: tuple[npt.NDArray[np.bool_], npt.NDArray[np.bool_]],
    ):
        self.targets, action, self.current_target = self.compute_action(
            self.targets, mask, self.current_target
        )

        return (action, self.current_target)

    @staticmethod
    def _action_func(
        current_target: np.int32 | None,
        targets: Array,
        surface_indexes: Array,
    ):
        # keep working on a target unless it has been removed from the attack surface
        return (
            choose_target(targets, surface_indexes)
            if current_target not in surface_indexes
            else (targets, current_target, False)
        )

    @staticmethod
    def _add_new_targets_func(targets: Array, new_targets: Array):
        # add new targets to the front of the list, so that the agent works on the latest targets first
        return np.concatenate([targets, new_targets])

```
```

File: malsim/sims/mal_sim_agent.py [hunk 3]
```diff
@@ -0,0 +1,87 @@
...
+
+
+class AgentType(Enum):
+    """Enum for agent types"""
+    ATTACKER = 'attacker'
+    DEFENDER = 'defender'
+
+
+@dataclass
+class MalSimAgent:
```
nkakouros, line 19: I think calling this `MalSimAgent` is confusing. This is a container for agent-related info. A name like `MalSimAgentState` or `MalSimAgentState` seems more representative. The current name adds one more "agent" to the actual agent classes and the maltoolbox attacker.

tests/sims/test_example_scenarios.py
nkakouros, line 1: Can you please explain how this test works? Why doesn't it use searcher agents? BFS is mentioned here but you seem to read the attackers from the toolbox?

malsim/sims/mal_sim_agent.py
mrkickling: Yes, this confusion is very annoying, and I think you are right about just calling it MalSimAgentState instead. Will change that.

tests/sims/test_example_scenarios.py
mrkickling: It doesn't use any searcher agents because I am not testing the searcher agents, but just the mal simulator functionality. I perform the steps one at a time with steps that I already know are available to the defender/attacker (as per the model and the entrypoints). The scenario file will probably help to see the entrypoints.

Yes, all attacker agents in the simulators in the MalSimulator must reference an Attacker object. This is so the simulator can know the entrypoints, reached attack steps and perform 'compromise' on nodes. This is nothing new in my PR though.

malsim/sims/mal_sim_agent.py
mrkickling: Fixed

malsim/agents/searchers.py
```diff
@@ -58,7 +58,12 @@ def get_next_action(
 
         return self.current_target
 
-    def _update_targets(self, action_surface):
+    def _update_targets(self, action_surface: list[AttackGraphNode]):
+
+        # action surface does not have a guaranteed order,
+        # so for the agent to be deterministic we need to sort
+        action_surface.sort(key=lambda n: n.id)
```
kasanari, line 65: If the order of the action surface can not be guaranteed (I personally think the simulator should handle such matters), then why typehint it as a list rather than a set?

malsim/cli.py
```diff
@@ -4,110 +4,65 @@
 import argparse
 import logging
 
-from .sims.mal_simulator import MalSimulator
-from .agents.keyboard_input import KeyboardAgent
-from .scenario import create_simulator_from_scenario
+from malsim.sims import MalSimulator
+from malsim.agents import DecisionAgent
+from malsim.scenario import create_simulator_from_scenario
```
nkakouros, line 9: Why do you revert to absolute imports?

malsim/cli.py
nkakouros, line 1: This really is the `__main__.py` file, right? How about renaming it?

malsim/sims/mal_sim_agent_state.py
```diff
@@ -0,0 +1,87 @@
...
+
+    @property
+    def action_surface(self) -> list[AttackGraphNode]:
+        return self._agent.action_surface
+
+    @property
+    def attacker_id(self) -> int:
+        assert isinstance(self._agent, MalSimAttackerState), \
+               "Only MalSimAttackers have property attacker_id"
+        return self._agent.attacker_id
```
nkakouros, line 87: A simpler implementation would be:

```suggestion
@dataclass
class MalSimAgentStateView(MalSimAgentState):
    """Read-only interface to MalSimAgentState."""
    def __post_init__(self) -> None:
        self.__setattr__ = self._freeze
        
    def _freeze(self, name, value) -> None:
        raise AttributeError("State view properties are read-only")
```

I haven't run this so it may need a few tweaks.

malsim/wrappers/malsim_vectorized_obs_logging_utils.py
nkakouros, line 1: This file is only ever imported by malsim_vectorized_obs_env.py. Why not merge it there already?

malsim/sims/__init__.py
nkakouros, line 1: This subpackage is called `sims`. But there is really only one simulator here, the MalSimulator. I suggest at least renaming the folder. Ideally, I would move all 4 files into one called simply mal_simulator.py. Or, just simulator.py for that matter. There is no other simulator in the whole package and if I understand correctly there will never be. The other things are environments, not simulators. So, if you want to import the simulator you would do the very natural `from malsim improt simulator` or `import malsim.simulator` or even `from malsim import Simulator`.

malsim/wrappers/base_classes.py
nkakouros, line 1: The namespace package `wrappers` could be renamed to the more descriptive `envs`. Each of the modules inside it could be renamed to `base_env.py`, `gym_env.py`, etc. Or plain `base.py`, `gym.py`, etc, to avoid the redundancy.

malsim/cli.py
mrkickling: No idea how that happened, thanks for noting.

malsim/agents/searchers.py
mrkickling: It will be changed into a set as soon as attackgraph nodes are hashable which is very soon

malsim/sims/__init__.py
mrkickling: Good point. It can probably all move into one file and it might not have to be a subpackage but can be just a file. But the class itself will continue to be called MalSimulator, since it is a MalSimulator and not just a Simulator.

malsim/wrappers/malsim_vectorized_obs_logging_utils.py
mrkickling: Will do that!

malsim/cli.py
mrkickling: That would match how it works in mal-toolbox yes so maybe it is worth to do

malsim/sims/mal_simulator.py
```diff
@@ -1,70 +1,40 @@
...
 
-import sys
 import copy
 import logging
-import functools
-from typing import Optional, TYPE_CHECKING
-import numpy as np
-
-from gymnasium.spaces import MultiDiscrete, Box, Dict
-from pettingzoo import ParallelEnv
```
nkakouros, line 11: All these dependencies were moved to other modules that are all optional to use. The main offering is the mal simulator. These dependencies should also be made [optional to install](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#dependencies-and-requirements) in pyproject.toml. Perhaps behind a `petting-zoo` and/or a `vectors` key(s).

malsim/agents/searchers.py
mrkickling: But you might be right that the simulator should give a sorted list already.

malsim/agents/searchers.py
```diff
@@ -1,137 +1,98 @@
...
-            targets.appendleft(current_target)
-            current_target = targets.pop()
+        self.settings = self.default_settings | agent_config
 
-        while current_target not in attack_surface:
-            if len(targets) == 0:
-                return None, True
+        self.rng = np.random.default_rng(
+            self.settings['seed'] or np.random.SeedSequence()
+        )
```
nkakouros, line 51: Consider using the `random` module here to drop the dependency on numpy for this core malsim module. For performance reasons, numpy may be re-introduced in the future for malsim, but I think it makes sense in this PR to have a clean state of what really needs to be used in what part.

malsim/sims/mal_simulator.py
```diff
@@ -75,1164 +45,378 @@ def __init__(
...
-
-                agent_info_str = self._format_info(infos[agent])
-                logger.debug(
-                    'Info for agent "%s":\n%s', agent, agent_info_str)
-
-        for agent in finished_agents:
-            self.agents.remove(agent)
+    def step(
+            self, actions: dict[str, list[AttackGraphNode]]
+        ) -> tuple[list[AttackGraphNode], list[AttackGraphNode]]:
```
nkakouros, line 348: I am confused as to how this is supposed to use. This method accepts a dict mapping agents to their selected actions. It returns a tuple with a single list of enabled actions for everyone. If I have two attackers shoudn't each one of them get a different list depending on their history of compromised steps?

malsim/sims/mal_simulator.py
```diff
@@ -75,1164 +45,378 @@ def __init__(
...
-                "model_edges_type": Box(
-                    0,
-                    num_lang_association_types,
-                    shape=(num_model_edges, ),
-                    dtype=np.int64,
-                ),  # instance model edge types
-            }
-        )
+        # Keep track on all 'living' agents sorted by order to step in
+        self.agents: list[str] = []
```
nkakouros, line 71: Why can't we use `self._agents_dict.keys()` ? They are also sorted (following insertion order).

tests/agents/test_searchers.py
```diff
@@ -0,0 +1,220 @@
...
+def test_breadth_first_traversal_simple():
+    """
+                    node1
+                      |
+                    node2
+                      |
+                    node3
+                      |
+                    node4
+    """
```
nkakouros, line 17: Can you explain how you test the breadth first behavior here? There is a single path to follow here (since you attach the attack at node1) so no matter what the attacker logic is, they will all follow the single path (unless they can redo stuff).

tests/run_demo.py
nkakouros, line 1: You deleted the single file that was the reference for how to use the simulator programmatically. I think it should be added back. It acts also as an integration (or whatever you call it) test that you could run in CI after pytest.

malsim/cli.py
```diff
@@ -4,110 +4,65 @@
...
-        total_reward_defender += rewards.get(defender_agent_id, 0) if defender_agent else 0
-        total_reward_attacker += rewards.get(attacker_agent_id, 0)
-
-        done |= terminated.get(attacker_agent_id, True) or truncated.get(attacker_agent_id, True)
+        for agent_dict in agents:
+            agent_name = agent_dict['name']
+            agent_state = sim.get_agent_state(agent_name)
+            total_rewards[agent_name] += agent_state.reward
+            if not agent_state.terminated and not agent_state.truncated:
+                all_agents_term_or_trunc = False
```
nkakouros, line 56: Looking at this code, I think the interface and code would be simplified if the agents were hashable and dicts like `total_rewards` or dicts internal to mal_simulator.py used them as keys, instead of using just their name. Since you use the agent name as its id (by using it as key to agent-related dicts), the hash function can return a hash of the agent name at its simplest. 

Doing the above may also remove the need for some of those `get_` methods as the user would just use dicts directly.

malsim/sims/mal_sim_agent_state.py
nkakouros: You can skip defining `_freeze` and use a lambda to save 2 lines.

malsim/wrappers/base_classes.py
mrkickling: Makes sense

malsim/sims/mal_simulator.py
mrkickling: Good point, I think the return values here is very much worth do discuss more. 
But yes, if you have more than one attacker/defender (this is not really something I have tested much), it would make sense to see which actually performed the action.

The thing though is that the attack graph is 'global', so if a defense node is enabled it will be enabled for all. If a node is compromised it will be compromised by only that attacker though. The multi agent case is something to look more into if we want to support it (which was the hope).

malsim/sims/mal_simulator.py
mrkickling: self.agents only contain the agents that are still 'alive' and not terminated (meaning they still have actions to take). However, I am not sure if we even care about that feature anymore.

tests/agents/test_searchers.py
mrkickling: You are right about that, the thought was to have a very basic test to see that the agent can receive action surface and make a decision, but it probably become too basic. It does test that the agent can make obvious decisions from an action surface though which is at least something.

tests/run_demo.py
mrkickling: I would say that the cli.py provides almost the same thing, but we can take it back for documentation purposes.

malsim/cli.py
mrkickling: That's a good idea I think, will look into that.

tests/run_demo.py
nkakouros: cli.py uses the scenario, right? I don't use a scenario and wanted to interact directly with API. cli.py is what I ended up looking at eventually, but it needed some trial and error. If you thing maintenance is too high, ignore the suggestion.

malsim/sims/mal_simulator.py
nkakouros: Regardless of the actual business logic, I think it makes sense to return a dict similar to the input/output of other methods (or even this one's input) to keep the interface consistent. What I realized is that you don't really care about the return value from step. You just step and then get the agent state and read the attack surface from there. So an option may be to return nothing from step. Or, you could just return the agents' state already to save another line of code for the users to get the agent state.

tests/run_demo.py
mrkickling: We could have a test that does it instead to not have maintenance be an issue

tests/agents/test_searchers.py
mrkickling: I do have another BFS test in the same file, but yes I will probably rename it

malsim/sims/mal_simulator.py
mrkickling: I think the return values do actually matter, you would want to know what was actually performed in the last step. This is clear if you look at the wrapper that uses the MalSimulator.

In the old simulator/the serialized obs wrapper it would return:
observations, rewards, terminations, truncations, infos (all dicts mapping from agent).

But the base simulator does not have to return that necessarily, for example observations does not really make sense. We do have `AgentState`s but it only contains action_surfaces, rewards and other agent info so just returning that would not be enough for someone using the MalSimulator to know which actions were actually performed in the last step.

Adding `performed_steps` to each `AgentState` could make `AgentState`s a better representation of each agents `observation`.

I am happy to change it to returning dicts that map performed steps per agent or to change it to returning `AgentState`s (if we add `performed_steps` for each agent state). The AgentState solution seems nicer if we can get all necessary information into them.

nkakouros: I am not following well. What are performed_steps? The steps the agent has already attempted or compromised in the past? Doesn't it already now that? Is this about using multiple agents and needing to know somehow what was performed by all? 

One thing is that the name `enabled_steps` could be improved. In the case of attacker agents, these are `performed_steps` or `compromised_steps`. Maybe the former captures both attacker and defender logic.

Second, isn't this information already available to the agents via the attack surface (the new additions and removals from it)? I am in favor of returning this information already here so that the agent does not have to compare the previous attack surface with the new one to see the difference that you can already provide them with. But in this case, it seems hard to get this right. For one, I don't think lots of agents will need this info often. Even the agent interface as is now does not allow them to receive this information. It would be useful for a BFS or DFS agent for example to get this info (what changed from the previous attack surface) and append or prepend the new steps in their ordered copy of the attack surface to help them implement their policy. They do it now internally anyway. But there is not way for them to receive that information. Second, the return from the `step()` function should be keyed per agent, otherwise the API becomes inconsistent and it will not support multiple attackers (even though they are not the scope of this PR). With a keyed return value you will need for each attacker to return their just performed step (they know that already) plus any disabled steps due to defenses). For the defender, you will to return the latter plus all the steps performed by the attackers. I think this becomes to messy as an API already to make sense, especially when we have the attack surfaces already that is great abstraction of all this and are per agent already.

malsim/sims/mal_sim_agent_state.py
nkakouros: Sth to note is that while the properties of the agent state are read-only, the values of the properties - if mutable - are not. For example, the action surface is not read only itself, the agent can modify it if they want. I think it would be a good idea to lock these down too, i.e. convert the action surface into a tuple of nodes instead of a list.

malsim/sims/mal_simulator.py
```diff
@@ -75,1164 +45,378 @@ def __init__(
...
+                    agent.name, node.full_name, node.id
+                )
 
-        for agent in self.agents:
-            agent_type = self.agents_dict[agent]["type"]
-            if  agent_type == "defender":
-                self._observe_defender(agent, performed_actions)
+                # Update attacker action surface
+                query.update_attack_surface_add_nodes(
+                    attacker, agent.action_surface, [node])
```
nkakouros, line 267: In these lines, the chosen step of an agent is checked first for traversability. If it's ok, it is compromised and the attack surface is updated. For this update to happen, the children of the just compromised step are checked for traversability and only those that pass the check are added to the surface. In the next timestep, the agent will choose one node from the attack surface which by definition will be traversable, making the first check on line 252 obsolete. 

Of course, we can argue that the simulator should check its inputs and not trust what the agents say, because they may be buggy, they may choose a step outside of the attack surface, etc. There are three options here:

- We do no checks and offload any requirement for upholding consistency to the agent implementation.
- See my comment [here](https://github.com/mal-lang/mal-simulator/pull/87#discussion_r1955304309). If we make the attack surface immutable (thus we can trust it and need not be concerned that the agent has altered it), we can then replace line 252 with a less expensive check of whether the attack step belonged to the attack surface.

malsim/agents/searchers.py
```diff
@@ -1,137 +1,98 @@
...
+        # action surface does not have a guaranteed order,
+        # so for the agent to be deterministic we need to sort
+        action_surface.sort(key=lambda n: n.id)
 
-    def compute_action_from_dict(self, observation: Dict[str, Any], mask: tuple):
-        new_targets, surface_indexes = get_new_targets(observation, self.targets, mask)
+        new_targets = [
+            step
+            for step in action_surface
+            if step not in self.targets and not step.is_compromised()
```
nkakouros, line 70: Would a compromised step ever end up in an agent's attack surface?

andrewbwm: Yes, nodes already compromised are still part of the attack surface since we do not know whether or not we want to allow them to be re-compromised or not. We could have this be a parameter for the `get_attack_surface` method, but I think just having it return a set as we were planning to do is cleaner.

malsim/sims/mal_simulator.py
andrewbwm: Yes, this is something @mrkickling and I discussed last week and we will probably have the `get_attack_surface` method return a set.

malsim/agents/searchers.py
nkakouros: This means that the attack surface grows as the simulation proceeds. At the end of a simulation where the agent has compromised almost everything the attack surface will be a copy basically of the attack graph. I think a better interface is needed here

andrewbwm: That is correct and has been brought up before. The attacker also tracks its reached attack steps which also end up covering the entire graph towards the end of simulations.

malsim/agents/decision_agent.py
```diff
@@ -0,0 +1,39 @@
...
+class PassiveAgent(DecisionAgent):
+    def __init__(self, *args, **kwargs):
+        ...
+
+    def get_next_action(
+        self,
+        agent: MalSimAgentStateView,
+        **kwargs
+    ) -> Optional[AttackGraphNode]:
+        ...
```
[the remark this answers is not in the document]
nkakouros: The DecisionAgent is an abstract class and can't be instantiated as is. We could make is non abstract but then you would use a DecisionAgent to do nothing (as a passive agent) which is confusing. We could remove the Decision Agent and use the Passive Agent as the base class for every other, but it would be counter intuitive as well to inherit from a passive agent and actually doing sth.

malsim/sims/mal_simulator.py
nkakouros: Isn't this resolved already?

malsim/sims/mal_sim_agent_state.py
```diff
@@ -0,0 +1,87 @@
...
+    action_surface: list[AttackGraphNode] = field(default_factory=list)
+
+    # Fields that tell if agent is 'dead' / disabled
+    truncated: bool = False
+    terminated: bool = False
+
+    # Fields that are not used by 'base' MalSimulator
+    # but can be useful in wrappers/envs
+    observation: dict = field(default_factory=dict)
+    info: dict = field(default_factory=dict)
```
nkakouros, line 40: Dataclasses can accept new properties on the fly, even if they are not defined beforehand. Maybe we should leave to the environments to defined these extra properties?

malsim/agents/searchers.py
mrkickling: The action surface is not necessarily ordered, it depends on how you look at it.
There could be an order (actions made available at a later time are added later in the action surface list).

If there is an order, we don't have a problem.
If there is not an order, and we still want a deterministic simulator when it comes to agent actions (which we definitely do), the action surface needs to be sorted somewhere.

Converting everything to sets as is being done now in the maltoolbox has made it harder to maintain deterministic simulations, this is why the sorting had to happen now in the agents. But as Jakob mentioned it might make sense that the action surface is ordered/sorted already in the simulator since we probably always want to be able to run deterministic simulations when a seed is given.

malsim/sims/mal_simulator.py
mrkickling: I think the simulator should definitely be responsible for checking that chosen actions are valid.

But, yes, having the action surface as a 'source of truth' probably makes sense, instead of having to check something multiple times.

malsim/sims/mal_sim_agent_state.py
mrkickling: Yes, these should be removed. Instead wrappers will store additional information in there own way, preferrably in there own data structures.

malsim/sims/mal_sim_agent_state.py
```diff
@@ -0,0 +1,87 @@
...
+    # Contains possible actions for the agent in the next step
+    action_surface: list[AttackGraphNode] = field(default_factory=list)
+
+    # Fields that tell if agent is 'dead' / disabled
+    truncated: bool = False
+    terminated: bool = False
+
+    # Fields that are not used by 'base' MalSimulator
+    # but can be useful in wrappers/envs
+    observation: dict = field(default_factory=dict)
```
andrewbwm, line 39: Why do MalSimAgents need an observation?

malsim/sims/mal_simulator.py
andrewbwm: Depending on the order of agent resolution in the simulator defenders may go before attackers in which case they can make the actions that the attackers have selected invalid. Does this make sense, do we want a more detailed example, or have I missed something?

nkakouros: @andrewbwm You are right. I missed that. But:
- if we make the attack surface immutable to be able to use it as the source of truth, and
- we replace the check on line 252 above with `if node in attacker.attack_surface`,

won't that cover all cases?

malsim/agents/searchers.py
nkakouros: In #98, I introduced a new variable `new_attack_surface` that keeps track of only the new nodes that became available in the just executed step. Whether to keep already compromised nodes in the attack surface is a good idea is an open topic. I am resolving this for now.

malsim/sims/mal_sim_agent_state.py
nkakouros: Done in #98

malsim/sims/mal_simulator.py
andrewbwm: 252 is a statement, I assume you mean the check on 249 that says `if query.is_node_traversable_by_attacker(node, attacker):`

Yes, in which case we can replace it with a check if the node is in the attack surface. What do you mean by making the attack surface immutable? The agent's own steps, plus those of the the defenders will always need to update it. But, your underlying assumption is correct, since it is updated as those steps are taken it already should only contain traversable nodes.

malsim/agents/decision_agent.py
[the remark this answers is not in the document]
mrkickling: Just chiming in: the PassiveAgent seems a bit useless, since you could just not provide an agent class at all and the result would be the same..

malsim/agents/searchers.py
mrkickling: It would be fine if it happens only if a seed is given I think.

Yes, I know we don't want the sorting to happen at all if we care about performance - but if we care about both performance AND determinism we should just have the action_surface be a list again created with deterministic order and never sorted after insertion. To have that, we would also need `attacker.reached_attack_steps` and `node.children` to be lists since they build up the attack surface, but that ship might have sailed..

malsim/agents/decision_agent.py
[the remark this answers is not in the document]
andrewbwm: Ok, I was very confused, I thought the PassiveAgent was a type of agent classes we wanted to have multiple variations for. There are two reasons for this. First, it's in the same file as the abstract class for no good reason, so my suggestion would be to move it to its own file called passive_agent. Second, it used `...` which I only ever use for classes I expect to override. It obviously works here, but my strong preference would be to have `get_next_action` explicitly `return None` for clarity. We're not paying storage costs by the letter.

malsim/sims/mal_simulator.py
nkakouros: > What do you mean by making the attack surface immutable?

I mean in the `MalSimAgentStateView`. See f93fd594dd9b5d7e0123ec8496560597b328cf93.

> We could technically make it even simpler and just check to see if it's unviable. It is the only way it could have become untraversable in the meantime.

Isn't `is_viable` a global node property and not scoped to an attacker? For example if an attack chooses a viable step that is not in its action surface, it will then be accepted and marked as compromised. Right?

malsim/agents/decision_agent.py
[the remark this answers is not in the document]
nkakouros: > my strong preference would be to have get_next_action explicitly return None

`...` returns None, although I don't object.

> move it to its own file called passive_agent

Or we could move all agents into the agents/__init_.py and simplify the file structure. We are using `__all__` in `__init__.py` anyway to expose them all at the top level under `malsim.agents` directly.

> the PassiveAgent seems a bit useless, since you could just not provide an agent class at all and the result would be the same.

I find it a bit useful because it gives you a common interface for you custom code to talk to and then swap agents on the fly. For example, I do in a project:

```python
        if timestep == x.start_step:
            l.info(f"Timestep {stepstamp}: Deploying {x.attacker} attacker.")
            attacker = attacker_agents[x.attacker](
                {
                    "randomize": True,
                    "attack_graph": env.attack_graph,
                    "wait_factor": 0 / 6,
                }
            )

        if timestep >= x.start_step and not state.terminated:
            # TODO check if timestep is still required to set here
            state.timestamp = timestep
            actions = _compute_action(attacker, state)
        else:
            # Emulate a passive attacker without creating one for the time steps
            # before an actual attacker agent (passive or other) is strictly
            # needed. Do that after the attacker has nothing left to do as well.
            actions = {"attacker": []}  # do nothing

        env.step(actions)
```

where `attacker = attacker_agents[x.attacker](...)` may be Passive, BFS, DFS, or other.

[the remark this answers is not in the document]
nkakouros: Also, having the passive agent as a class allows easy use via scenario files.

malsim/sims/mal_simulator.py
andrewbwm: > > What do you mean by making the attack surface immutable?
> 
> I mean in the `MalSimAgentStateView`. See [f93fd59](https://github.com/mal-lang/mal-simulator/commit/f93fd594dd9b5d7e0123ec8496560597b328cf93).

It's still not entirely clear to me what making the action surface immutable entails. I haven't internalized this redesign, maybe it will become clear to me down the line, but the action surface is mutable and it does change. If we lock it in place after certain operations that's fine, but I'm not sure what that has to do with anything. I suspect that I haven't reached that part of the redesign, so maybe we can leave this for now and I can get back to it when I'm done.

> 
> > We could technically make it even simpler and just check to see if it's unviable. It is the only way it could have become untraversable in the meantime.
> 
> Isn't `is_viable` a global node property and not scoped to an attacker? For example if an attack chooses a viable step that is not in its action surface, it will then be accepted and marked as compromised. Right?

Yes, the `is_viable` property is a node global, it applies to all attackers. The agent is expected  to only pick things from its attack surface though, the most logical thing would actually be to check both.

However, as I was thinking about this I just realized that we have two logical mistakes in how we handle both the attack surface and is_traversable right now.

### Attack Surface Mistake
The attack surface update is potentially incorrect because it will remove unviable nodes from the attack surface, but it will not remove viable nodes that no longer have any parents in the attack surface.

So, if they attacker got to `node B`(which is of type `or`) via `node A`. And then through defender actions `node A` becomes unviable. If `node B` also had `node C` as a parent that is still viable, but the attacker had not reached it, it would still be left in the attack surface as it would still be viable.

This is something that I do think we've discussed before because it isn't implicitly incorrect. An attacker reading a password still means they can use it later even if the file in the meantime has become encrypted before they tried to log in with that user, for example.

### Is Traversable Mistake
If an `and`  node is checked the method will check to see if all of its necessary parents are already compromised by the attacker, which is correct. However, if the node is of type `or` it simply returns `True`, it doesn't check to see if at least one parent is compromised at this point. This one definitely feels like a mistake and something we should just fix.

malsim/agents/decision_agent.py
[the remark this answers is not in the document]
andrewbwm: I still think we want it too, it's a convenient feature. I just want it more obvious in what it does and what it is meant for.

malsim/sims/mal_simulator.py
nkakouros: > ### Attack Surface Mistake
> 
> The attack surface update is potentially incorrect because it will remove unviable nodes from the attack surface, but it will not remove viable nodes that no longer have any parents in the attack surface.
> 
> So, if they attacker got to `node B`(which is of type `or`) via `node A`. And then through defender actions `node A` becomes unviable. If `node B` also had `node C` as a parent that is still viable, but the attacker had not reached it, it would still be left in the attack surface as it would still be viable.
> 
> This is something that I do think we've discussed before because it isn't implicitly incorrect. An attacker reading a password still means they can use it later even if the file in the meantime has become encrypted before they tried to log in with that user, for example.

I think this is sth that should be solved at the language level, because it depends on the semantics of the attack step (node B in your example). If it is a service, for example, that you can access directly or through a proxy, If you access it directly and a firewall is enabled to block direct access, you should not still be logged in. I can't tell if this can be solved with current MAL, i.e., by carefully using the boolean logic of the steps or if we need an annotation to explicitly dictate the behavior.

malsim/sims/mal_simulator.py
```diff
@@ -75,1173 +45,375 @@ def __init__(
...
-            self._observe_and_reward(
-                performed_actions,
-                prevented_attack_steps
-            )
-        )
+        if all_attackers_terminated:
+            # Terminate all agents if all attackers are terminated
+            logger.info("All attackers are terminated")
+            for agent in self._agents_dict.values():
+                agent.terminated = True
```
nkakouros, line 393: Is this needed? Looking at lines 374-375, `all_attackers_terminated` will be set to False if any of the attackers is not terminated. It will be True if they are all terminated. In this latter case, this loop will run, terminating all attackers, but they are already terminated.

malsim/sims/mal_simulator.py
```diff
@@ -75,1173 +45,375 @@ def __init__(
...
+
+        agents_to_remove = set()
+        for agent_name in self.agents:
+            agent = self._agents_dict[agent_name]
+            if agent.terminated or agent.truncated:
+                logger.info("Removing agent %s", agent.name)
+                agents_to_remove.add(agent.name)
+
+        for agent in agents_to_remove:
+            self.agents.remove(agent)
```
nkakouros, line 409:
```suggestion
        agents_to_remove = set()
        for agent_name in self.agents:
            agent = self._agents_dict[agent_name]
            if agent.terminated or agent.truncated:
                logger.info("Removing agent %s", agent.name)
                self.agents.remove(agent)
```

This should work the same right?

malsim/sims/mal_simulator.py
mrkickling: It should only be needed for terminating the defender agents, since if the attackers are terminated the defenders should also be terminated. At least this is how it used to work.

malsim/sims/__init__.py
mrkickling: I will push this change now, you might want to rebase your other branches on this.

malsim/sims/mal_simulator.py
mrkickling: Hm, do you mean we can be removing items from the list we are iterating over? I don't think that would work.

nkakouros: Ah right. I updated the suggestion to loop over a copy of the list.

mrkickling: Yes, I like that, since it is just a list of strings and not expected to be very large this works well and is fewer lines of code. Will implement.

malsim/agents/searchers.py
```diff
@@ -1,137 +1,98 @@
...
-            if len(targets) == 0:
-                return None, True
+        self.rng = np.random.default_rng(
+            self.settings['seed'] or np.random.SeedSequence()
+        )
 
-            current_target = targets.pop()
+    def get_next_action(
+        self, agent: MalSimAgentStateView, **kwargs
+    ) -> Optional[AttackGraphNode]:
```
nkakouros, line 55: The returned action should be wrapped in a dict like:

```
selected_action = agent.get_next_action(...)
action = {"attacker": action}
sim.step(action)
```
Would it make sense to have this method return the action wrapped already? It does not look elegant to me to do that, but neither does manually wrapping the action in the dict.

mrkickling: Not sure which way I prefer more, maybe we can let it stay this way for the meantime.

malsim/wrappers/gym_wrapper.py
```diff
@@ -23,13 +26,28 @@ def __init__(self, scenario_file: str, **kwargs) -> None:
         self.render_mode = kwargs.pop('render_mode', None)
 
         # Create a simulator from the scenario given
-        self.sim, _ = create_simulator_from_scenario(scenario_file, **kwargs)
+        attack_graph, agents = load_scenario(scenario_file, **kwargs)
+        self.sim = MalSimVectorizedObsEnv(MalSimulator(attack_graph))
```
nkakouros, line 30: I find it confusing that self.sim actually holds an environment, or at least that's what its name says, `MalSimVectorizedObsEnv`. If this is truly an environment, then `self.sim` should be renamed. Could it be that we actually want custom gym envs to subclass the malsim vectorized env as well? In any case, it is a bit messy right now I think. From a gym env, you could do `self.sim.sim.<MalSimulator_thing>` which is confusing.

malsim/sims/mal_simulator.py
nkakouros: Done in #98

malsim/sims/mal_simulator.py
nkakouros: About this check:

> Yes, the is_viable property is a node global, it applies to all attackers. The agent is expected to only pick things from its attack surface though, the most logical thing would actually be to check both.

see https://github.com/mal-lang/mal-simulator/pull/98#issuecomment-2676344414.

The traversable mistake is fixed. The attack surface mistake is open for discussion, I created #100.

malsim/cli.py
mrkickling: Leaving this for later discussion

nkakouros: Do we have issues to track all these for-later discussions?

PR #90, opened 2025-01-13
Rewrite agents

nkakouros:
Rewrites the bundled agents.

- Simplified code and removed repetition
- Tries chosen attack step until it succeeds (i.e. until TTC is depleted)

### DIFF
--- a/malsim/agents/searchers.py
+++ b/malsim/agents/searchers.py
@@ -1,137 +1,79 @@
 import logging
+import re
 
-from collections import deque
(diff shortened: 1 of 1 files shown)


kasanari:
While I have some other objections to the changes here in terms of the style and changed functionality, my main note is that any change of the agent code is very precarious, since there are no tests in the current simulator that checks the edge cases of agent interaction. What is especially important is the interaction between the defender and attacker, such as when an attack step is removed from the attack surface by an defense step.

mrkickling:
Should this one be closed?

PR #94, opened 2025-01-24
Adapt to maltoolbox 0.2.0

mrkickling:
- Adapt the current (old) mal simulator to the new mal-toolbox 0.2.0 with changes to the language graph.
- Convert all test models to new model format.
- Make sure tests pass

The intention is to merge this to main and create a release before we are done with #87, just to support maltoolbox 0.2.0 as soon as possible in the simulator.

### DIFF
--- a/malsim/scenario.py
+++ b/malsim/scenario.py
@@ -97,8 +97,8 @@ def _validate_scenario_property_rules(
 
     # a way to lookup attack steps for asset types
     asset_type_step_names = {
-        asset_type.name: [a.name for a in asset_type.attack_steps]
-        for asset_type in graph.lang_graph.assets
+        asset_type_name: [step for step in asset_type.attack_steps]
+        for asset_type_name, asset_type in graph.lang_graph.assets.items()
     }
 
     if rules is None:
--- a/malsim/sims/mal_simulator.py
+++ b/malsim/sims/mal_simulator.py
@@ -159,15 +159,26 @@ class MalSimulator(ParallelEnv):
             right_field = getattr(assoc, right_field_name)
             for left_asset in left_field:
                 for right_asset in right_field:
+
                     observation["model_edges_ids"].append(
                         [
                             self._model_asset_id_to_index[left_asset.id],
                             self._model_asset_id_to_index[right_asset.id]
                         ]
                     )
+
+                    association_type = (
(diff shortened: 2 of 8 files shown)


mrkickling:
The test that failed is because the order of choices for the BFS attacker are not the same anymore. Why this happens is worth looking into, but probably it is just because the observations are built up differently. I will compare two attack graphs (one from mal-toolbox 0.1.0 and one from 0.2.0) and see if that gives any clues.

If nothing seems fishy I will just rewrite the test to match the new order.

mrkickling:
The main thing I noticed is that the IDs are not the same in the new 0.2.0 attack graph compared to the old 0.1.0 attack graph (created from the same model). Does this seem normal  @andrewbwm ? Other than that the two attack graphs seem identical (I have not checked thoroughly) so I will just rewrite the test on monday and then put this test in review.

andrewbwm:
> The main thing I noticed is that the IDs are not the same in the new 0.2.0 attack graph compared to the old 0.1.0 attack graph (created from the same model). Does this seem normal @andrewbwm ? Other than that the two attack graphs seem identical (I have not checked thoroughly) so I will just rewrite the test on monday and then put this test in review.

That is normal because attack steps are probably created in a slightly different order now, but it shouldn't really matter. Also, you don't need to adjust their ids between versions, there is no expectation about the ordering of ids. They just need to be unique.

It's also a good idea if we have the tests not really depend too much on ids, but rather on names. Not sure which test this is but we can have a look at it if it isn't obvious how to rewrite it to use names.

mrkickling:
> > The main thing I noticed is that the IDs are not the same in the new 0.2.0 attack graph compared to the old 0.1.0 attack graph (created from the same model). Does this seem normal @andrewbwm ? Other than that the two attack graphs seem identical (I have not checked thoroughly) so I will just rewrite the test on monday and then put this test in review.
> 
> That is normal because attack steps are probably created in a slightly different order now, but it shouldn't really matter. Also, you don't need to adjust their ids between versions, there is no expectation about the ordering of ids. They just need to be unique.
> 
> It's also a good idea if we have the tests not really depend too much on ids, but rather on names. Not sure which test this is but we can have a look at it if it isn't obvious how to rewrite it to use names.

Good, good!
I will use the name instead, and it is quite obvious how to convert it.

PR #97, opened 2025-02-13
Upgrade to maltoolbox 0.3.0

mrkickling:


malsim/scenario.py
```diff
@@ -331,9 +331,15 @@ def create_simulator_from_scenario(
...
     )
 
+    # This version only supports one defender and one attacker
     for agent_id, agent_info in conf['agents'].items():
         if agent_info['type'] == "attacker":
-            sim.register_attacker(agent_id, 0)
+            all_attackers = list(attack_graph.attackers.values())
+            assert len(all_attackers) == 1, (
+                "You have defined more than one attacker,",
+                "can not decide which one belongs to agent in simulator"
```
andrewbwm, line 340: "cannot" and maybe make this a log error instead. To me I'd leave this as a warning and just still take the first attacker, but it's not a meaningful discussion anyway. So, just go with your gut.

malsim/sims/mal_simulator.py
```diff
@@ -103,42 +103,63 @@ def create_blank_observation(self, default_obs_state=-1):
...
-                          str(step.asset.type + ":" + step.name)]
-                          for step in self.attack_graph.nodes],
+            "asset_type": [self._asset_type_to_index[step.lg_attack_step.asset.name]
+                           for step in ordered_nodes],
+            "asset_id": [step.model_asset.id
+                         for step in ordered_nodes],
+            "step_name": [
+                self._step_name_to_index.get(
+                    str(step.lg_attack_step.asset.name + ":" + step.name)
+                ) for step in ordered_nodes],
```
andrewbwm, line 130: I'm not sure which one is faster sorting the dictionary and then having these lists created like they are right now or just iterating through the keys and appending to the lists one element at a time.

malsim/sims/mal_simulator.py
```diff
@@ -536,10 +541,14 @@ def _create_mapping_tables(self):
         """Create mapping tables"""
         logger.debug("Creating and listing mapping tables.")
 
+        # Since nodes are stored as sets and we need deterministic
+        # observations in the simulator, we need to convert them
+        # to a list and sort them
+
```
andrewbwm, line 547: Outdated comment?

mrkickling: Yes, thanks.

malsim/scenario.py
```diff
@@ -331,9 +331,15 @@ def create_simulator_from_scenario(
         **kwargs
     )
 
+    # This version only supports one defender and one attacker
     for agent_id, agent_info in conf['agents'].items():
         if agent_info['type'] == "attacker":
-            sim.register_attacker(agent_id, 0)
+            all_attackers = list(attack_graph.attackers.values())
```
nkakouros, line 337: no need to cast to list, len works with `dict_values` as well.

malsim/sims/mal_simulator.py
```diff
@@ -106,39 +106,52 @@ def create_blank_observation(self, default_obs_state=-1):
...
+                ordered_children = list(attack_step.children)
+                ordered_children.sort(key=lambda n: n.id)
+                for child in ordered_children:
                     observation["attack_graph_edges"].append(
-                        [self._id_to_index[child.id],
-                            self._id_to_index[attack_step.id]]
+                        [
+                            self._id_to_index[child.id],
+                            self._id_to_index[attack_step.id]
+                        ]
```
nkakouros, line 154: These two loops can be combined into one and avoid sorting twice.

tests/test_mal_simulator.py
```diff
@@ -777,7 +864,8 @@ def verify_attacker_obs_state(
     )
     sim.reset()
 
-    attacker = sim.attack_graph.attackers[0]
+    all_attackers = list(sim.attack_graph.attackers.values())
+    attacker = all_attackers[0]
```
andrewbwm, line 868: If we just want the first attacker we can use `next(iter(sim.attack_graph.attackers.values()))`

andrewbwm: Or, if we know its id, we can just fetch it from the dictionary instead. Might look cleaner.

tests/testdata/models/network_app_data_model.yml
```diff
@@ -0,0 +1,61 @@
...
+    associated_assets: {}
+    extras:
+      position:
+        x: -1617.3333333333333
+        y: -1299.3333333333333
+    name: SoftwareVuln
+    type: SoftwareVulnerability
+attackers: {}
+metadata:
+  MAL-Toolbox Version: 0.2.0
```
andrewbwm, line 56: This says that it was created using the 0.2.0 version. Did you manually update the model to fit the current format? Or did you update the mal-gui?

tests/testdata/models/run_demo_model.json
```diff
@@ -10,100 +10,45 @@
     "assets": {
         "0": {
             "name": "OS App",
-            "type": "Application"
+            "type": "Application",
+            "associated_assets": {}
```
andrewbwm, line 14: None of these assets have their associations left anymore. Why do we want this model then?

malsim/__init__.py
```diff
@@ -31,8 +31,10 @@
 
 __all__ = ("AttackerEnv", "DefenderEnv", "register_envs")
 
+
+# TODO: Make sure logging dir exists and make it configurable (or use same as maltoolbox)
 formatter = logging.Formatter('%(asctime)s %(name)-12s %(levelname)-8s %(message)s', datefmt='%m-%d %H:%M')
-file_handler = logging.FileHandler('tmp/malsim_log.txt', mode='w')
+file_handler = logging.FileHandler('logs/malsim_log.txt', mode='w')
```
nkakouros, line 37: Perhaps we could have a single config file for all these tools, not just one specific to the mal toolbox. The mal toolbox will still load it and the config will be importable from the toolbox to the simulator/gui/etc. It's just a conceptual thing I think and a matter of how we name the config file. Sth like `malconfig.yml` or `maltools.yml`.

malsim/scenario.py
mrkickling: True!

tests/test_mal_simulator.py
mrkickling: I use the next iter approach now instead, I agree it is better

tests/testdata/models/network_app_data_model.yml
mrkickling: I need to dig into this, could be some issue with the maltoolbox upgrader, but not sure actually.

mrkickling: Fixed

PR #98, opened 2025-02-18
Implements suggestions made in PR #87

nkakouros:
I implemented here most of my suggestions in #87. See the commit messages for what they do. They tests fail because the mal-toolbox version used here is missing https://github.com/mal-lang/mal-toolbox/pull/121. When this gets merged, the tests will also pass. The suggestions I did not implement have to do with renaming folders/files as that would make following the changes harder. There are also a couple TODOs added imin the code about things not clear to me. Depending on what the answer is to those TODOs the code may be possible to simplify in some parts.

One of the things this PR does is change what `def step` returns. It now returns three things:

1. actioned_nodes, i.e. the nodes that each attacker/defender agent has compromised/enabled. This is a dict and each agent receives only what is relevant to them
1. new_attack_surface, this is also a dict and contains the new steps that became available in the latest timestep to each agent.  This is set only for attackers, since defenders start with an defense surface that contains all defenses and defenses get removed as they are enabled, thus no new defense surface ever becomes available. 
1. disabled nodes: the nodes that became disabled in the last timestep due to defender actions. This is a simple set that is common to all agents.

1 was changed from what is currently in #87 to be a dict, 2 is a new addition and 3 is left as is.

The advantage of having 2 available is that this info can be passed to the agents and they don't have to calculate it themselves.

While the above works, I think it would be better if we extended the `MalSimAgentState` class with three new keys to hold the three bits of information above:

```
@dataclass
class MalSimAgentState:
    """Stores the state of an agent in the simulator"""

    # Identifier of the agent, used in MalSimulator for lookup
    name: str
    type: AgentType

    # Contains current agent reward in the simulation
    # Attackers get positive rewards, defenders negative
    reward: int = 0

    # Contains possible actions for the agent in the next step
    action_surface: set[AttackGraphNode] = field(default_factory=set)
    new_action_surface: ....
    actioned_nodes: ....
    disabled_nodes: ....

    # Fields that tell if agent is 'dead' / disabled
    truncated: bool = False
    terminated: bool = False

    # Fields that are not used by 'base' MalSimulator
    # but can be useful in wrappers/envs
    observation: dict = field(default_factory=dict)
    info: dict = field(default_factory=dict)
```

I think it is natural to do this as these change on every time step. And if we accept action_surface to be a good fit for this class, then these three surely are as well.

If the above happens (trivial), then the return value of `step()` could change to sth simpler and more intuititve, i.e. a dict like:

```
{
    "attacker1": MalSimAgentState(....for-attacker1...),
    "defender1": MalSimAgentState(....for-defender1....),
```

### DIFF
--- a/.github/workflows/test-pytest.yml
+++ b/.github/workflows/test-pytest.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install .
+          pip install ".[ml]"
           pip install pytest
       - name: Test with pytest
         run: |
--- a/malsim/mal_simulator.py
+++ b/malsim/mal_simulator.py
@@ -4,7 +4,8 @@ import copy
 from dataclasses import dataclass, field
 import logging
 from enum import Enum
-from typing import Optional
+from types import MappingProxyType
+from typing import TYPE_CHECKING, Any, Generic, Optional, TypeVar
 
 from maltoolbox import neo4j_configs
 from maltoolbox.ingestors import neo4j
@@ -13,6 +14,10 @@ from maltoolbox.attackgraph.analyzers import apriori
 
 ITERATIONS_LIMIT = int(1e9)
 logger = logging.getLogger(__name__)
+
+
+if TYPE_CHECKING:
+    from maltoolbox.attackgraph import AttackGraphNode
 
 
 class AgentType(Enum):
@@ -33,21 +38,26 @@ class MalSimAgentState:
     # Attackers get positive rewards, defenders negative
     reward: int = 0
 
-    # Contains possible actions for the agent in the next step
-    action_surface: list[AttackGraphNode] = field(default_factory=list)
+    # Contains possible actions for the agent on the next step
+    action_surface: set[AttackGraphNode] = field(default_factory=set)
+
+    # Contains possible actions that became available on the last step
+    step_action_surface: set[AttackGraphNode] = field(default_factory=set)
+
+    # Contains the steps performed successfully in the last step
+    step_compromised_nodes: set[AttackGraphNode] = field(default_factory=set)
+
+    # Contains nodes that defender actions made unviable in the last step
+    step_disabled_nodes: set[AttackGraphNode] = field(default_factory=set)
 
     # Fields that tell if agent is 'dead' / disabled
     truncated: bool = False
     terminated: bool = False
 
-    # Fields that are not used by 'base' MalSimulator
-    # but can be useful in wrappers/envs
-    observation: dict = field(default_factory=dict)
-    info: dict = field(default_factory=dict)
-
 
 class MalSimAttackerState(MalSimAgentState):
     """Stores the state of an attacker in the simulator"""
+
     def __init__(self, name: str, attacker_id: int):
         super().__init__(name, AgentType.ATTACKER)
         self.attacker_id = attacker_id
@@ -55,44 +65,52 @@ class MalSimAttackerState(MalSimAgentState):
 
 class MalSimDefenderState(MalSimAgentState):
     """Stores the state of a defender in the simulator"""
+
+    # Contains defense steps successfully enabled by the defender in the last step
+    step_enabled_defenses: set[AttackGraphNode] = field(default_factory=set)
+
     def __init__(self, name: str):
         super().__init__(name, AgentType.DEFENDER)
 
 
-class MalSimAgentStateView:
+T = TypeVar("T", bound=MalSimAgentState)
+
+
+class MalSimAgentStateView(Generic[T]):
     """Read-only interface to MalSimAgentState."""
-    def __init__(self, agent: MalSimAgentState):
+
+    _frozen = False
+
+    def __init__(self, agent: T):
         self._agent = agent
+        self._frozen = True
 
-    @property
-    def name(self) -> str:
-        return self._agent.name
+    def __setattr__(self, key, value) -> None:
+        if self._frozen:
+            raise AttributeError("Cannot modify agent state view")
+        self.__dict__[key] = value
 
-    @property
-    def type(self) -> AgentType:
-        return self._agent.type
+    def __delattr__(self, key) -> None:
+        if self._frozen:
+            raise AttributeError("Cannot modify agent state view")
+        super().__delattr__(key)
 
-    @property
-    def reward(self) -> int:
-        return self._agent.reward
+    def __getattr__(self, attr) -> Any:
+        """Return read-only version of proxied agent's properties."""
+        value = getattr(self._agent, attr)
 
-    @property
-    def truncated(self) -> bool:
-        return self._agent.truncated
+        if isinstance(value, dict):
+            return MappingProxyType(value)
+        if isinstance(value, list):
+            return tuple(value)
+        if isinstance(value, set):
+            return frozenset(value)
 
-    @property
-    def terminated(self) -> bool:
-        return self._agent.terminated
+        return value
 
-    @property
-    def action_surface(self) -> list[AttackGraphNode]:
-        return self._agent.action_surface
-
-    @property
-    def attacker_id(self) -> int:
-        assert isinstance(self._agent, MalSimAttackerState), \
-               "Only MalSimAttackers have property attacker_id"
-        return self._agent.attacker_id
+    def __dir__(self):
+        """Dynamically resolve attribute names for REPL autocompletion."""
+        return list(vars(self._agent).keys()) + ["_agent", "_frozen"]
 
 
 @dataclass
@@ -112,6 +130,7 @@ class MalSimulatorSettings():
     # - Defender only sees the status of nodes changed in the current step
     cumulative_defender_obs: bool = True
 
+
 class MalSimulator():
     """A MAL Simulator that works on the AttackGraph
 
@@ -129,8 +148,6 @@ class MalSimulator():
     ):
         """
         Args:
-            lang_graph                  -   The language graph to use
-            model                       -   The model to use
             attack_graph                -   The attack graph to use
             max_iter                    -   Max iterations in simulation
             prune_unviable_unnecessary  -   Prunes graph if set to true
@@ -148,28 +165,22 @@ class MalSimulator():
 
         # Initialize all values
         self.attack_graph = attack_graph
-        self.lang_graph = attack_graph.lang_graph
-        self.model = attack_graph.model
 
         self.sim_settings = sim_settings
-        self.max_iter = max_iter # Max iterations before stopping simulation
+        self.max_iter = max_iter  # Max iterations before stopping simulation
         self.cur_iter = 0        # Keep track on current iteration
 
         # Keep track on all registered agent states
-        self._agents_dict: dict[str, MalSimAgentState] = {}
+        self.agents: dict[str, MalSimAgentState] = {}
 
         # Keep track on all 'living' agents sorted by order to step in
-        self.agents: list[str] = []
-
-        # Keep track on all (dead and alive) agents sorted by order to step in
-        # (Used when resetting)
-        self.possible_agents: list[str] = []
+        self.alive_agents: dict[str, MalSimAgentState] = {}
 
     def reset(
-            self,
-            seed: Optional[int] = None,
-            options: Optional[dict] = None
-        ) -> dict[str, MalSimAgentState]:
+        self,
+        seed: Optional[int] = None,
+        options: Optional[dict] = None
+    ) -> dict[str, MalSimAgentStateView]:
         """Reset attack graph, iteration and reinitialize agents"""
 
         logger.info("Resetting MAL Simulator.")
@@ -180,7 +191,7 @@ class MalSimulator():
         # Reset agents
         self._reset_agents()
 
-        return self._agents_dict
+        return self.agent_states
 
     def _init_agent_rewards(self):
         """Give rewards for pre-enabled attack/defense steps"""
@@ -189,7 +200,7 @@ class MalSimulator():
             if not node_reward:
                 continue
 
-            for agent in self._agents_dict.values():
+            for agent in self.agents.values():
 
                 if agent.type == AgentType.ATTACKER:
                     attacker = self.attack_graph.attackers[agent.attacker_id]
@@ -205,18 +216,12 @@ class MalSimulator():
 
     def _init_agent_action_surfaces(self):
         """Set agent action surfaces according to current state"""
-        for agent in self._agents_dict.values():
+        for agent in self.agents.values():
             if agent.type == AgentType.ATTACKER:
-                # Get the Attacker object
-                attacker = \
-                    self.attack_graph.attackers[agent.attacker_id]
-
-                # Get current action surface
-                agent.action_surface = \
-                    query.get_attack_surface(attacker)
+                attacker = self.attack_graph.attackers[agent.attacker_id]
+                agent.action_surface = query.calculate_attack_surface(attacker)
 
             elif agent.type == AgentType.DEFENDER:
-                # Get current action surface
                 agent.action_surface = \
                     query.get_defense_surface(self.attack_graph)
             else:
@@ -225,10 +230,12 @@ class MalSimulator():
     def _reset_agents(self):
         """Reset agent rewards and action surfaces"""
 
-        # Revive dead agents
-        self.agents = copy.deepcopy(self.possible_agents)
+        self.alive_agents = dict(sorted(
+            self.agents.items(),
+            key=lambda agent: agent[1].type == AgentType.ATTACKER,
+        ))
 
-        for agent in self._agents_dict.values():
+        for agent in self.agents.values():
             # Reset agent reward
             agent.reward = 0
 
@@ -243,20 +250,17 @@ class MalSimulator():
         """Register a mal sim agent"""
 
         logger.info('Registering agent "%s".', agent)
-        assert agent.name not in self._agents_dict, \
+        assert agent.name not in self.agents, \
             f"Duplicate agent named {agent.name} not allowed"
 
-        if agent.type == AgentType.DEFENDER:
-            # Defender is first in list so it can pick
-            # actions before attacker when step performed
-            self.agents.insert(0, agent.name)
-            self.possible_agents.insert(0, agent.name)
-        elif agent.type == AgentType.ATTACKER:
-            # Attacker goes last
-            self.agents.append(agent.name)
-            self.possible_agents.append(agent.name)
+        self.agents[agent.name] = agent
 
-        self._agents_dict[agent.name] = agent
+        # Defender is first in list so it can pick
+        # actions before attacker when step performed
+        self.alive_agents = dict(sorted(
+            self.agents.items(),
+            key=lambda agent: agent[1].type == AgentType.ATTACKER,
+        ))
 
     def register_attacker(self, name: str, attacker_id: int):
         """Register a mal sim attacker agent"""
@@ -268,31 +272,22 @@ class MalSimulator():
         agent_state = MalSimDefenderState(name)
         self._register_agent(agent_state)
 
-    def get_agent_state(self, name: str) -> MalSimAgentStateView:
-        """Return read only agent state for agent with given name"""
-
-        assert name in self._agents_dict, (
-            f"Agent with name '{name}' does not exist")
-        agent = self._agents_dict[name]
-        return MalSimAgentStateView(agent)
-
-    def get_agents(self) -> list[MalSimAgentStateView]:
+    @property
+    def agent_states(self) -> list[MalSimAgentStateView]:
         """Return read only agent state for all dead and alive agents"""
-        return [self.get_agent_state(agent) for agent in self.possible_agents]
+        return {name: MalSimAgentStateView(agent) for name, agent in self.agents.items()}
 
     def _get_attacker_agents(self) -> list[MalSimAttackerState]:
         """Return list of mutable attacker agent states"""
-        return [a for a in self._agents_dict.values()
-                if a.type == AgentType.ATTACKER]
+        return [a for a in self.agents.values() if a.type == AgentType.ATTACKER]
 
     def _get_defender_agents(self) -> list[MalSimDefenderState]:
         """Return list of mutable defender agent states"""
-        return [a for a in self._agents_dict.values()
-                if a.type == AgentType.DEFENDER]
+        return [a for a in self.agents.values() if a.type == AgentType.DEFENDER]
 
     def _disable_attack_steps(
-            self, attack_steps_to_disable: list[AttackGraphNode]
-        ):
+        self, attack_steps_to_disable: set[AttackGraphNode]
+    ):
         """Disable nodes for each attacker agent
 
         For each compromised attack step uncompromise the node, disable its
@@ -316,18 +311,19 @@ class MalSimulator():
                     attacker.undo_compromise(unviable_node)
 
     def _attacker_step(
-            self, agent: MalSimAttackerState, nodes: list[AttackGraphNode]
-        ) -> list[AttackGraphNode]:
+        self, agent: MalSimAttackerState, nodes: list[AttackGraphNode]
+    ) -> tuple[set[AttackGraphNode], set[AttackGraphNode]]:
         """Compromise attack step nodes with attacker
 
         Args:
         agent - the agent to compromise nodes with
         nodes - the nodes to compromise
 
-        Returns: list of attack steps nodes that were compromised
+        Returns: tuple of set of attack steps nodes that were compromised and
+                 the new nodes that became available in the attack surface
         """
 
-        enabled_nodes = []
+        compromised_nodes = set()
         attacker = self.attack_graph.attackers[agent.attacker_id]
 
         for node in nodes:
@@ -349,19 +345,23 @@ class MalSimulator():
                 agent.reward += node_reward
                 for d_agent in self._get_defender_agents():
                     d_agent.reward -= node_reward
-                enabled_nodes.append(node)
+                compromised_nodes.add(node)
 
                 logger.info(
                     'Attacker agent "%s" compromised "%s"(%d).',
                     agent.name, node.full_name, node.id
                 )
-
-                # Update attacker action surface
-                query.update_attack_surface_add_nodes(
-                    attacker, agent.action_surface, [node])
             else:
                 logger.warning("Attacker could not compromise %s",
-                            node.full_name)
+                               node.full_name)
+
+        # Update attacker action surface
+        new_attack_surface = query.calculate_attack_surface(
+            attacker, from_nodes=compromised_nodes, skip_compromised=True
+        )
+
+        # TODO: why extend?
+        agent.action_surface |= new_attack_surface
 
         # Terminate attacker if it has nothing left to do
         terminate = True
@@ -371,22 +371,23 @@ class MalSimulator():
                 break
         agent.terminated = terminate
 
-        return enabled_nodes
+        return compromised_nodes, new_attack_surface
 
     def _defender_step(
-            self, agent: MalSimDefenderState, nodes: list[AttackGraphNode]
-        ) -> tuple[list[AttackGraphNode], list[AttackGraphNode]]:
-        """Enable defense step nodes with defender
+        self, agent: MalSimDefenderState, nodes: list[AttackGraphNode]
+    ) -> tuple[set[AttackGraphNode], set[AttackGraphNode]]:
+        """Enable defense step nodes with defender.
 
         Args:
         agent - the agent to activate defense nodes with
         nodes - the defense step nodes to enable
 
-        Returns: list of defense steps nodes that were enabled
+        Returns: tuple of defense steps that were enabled and attack steps that
+                 became unviable due to enabled defenses
         """
 
-        attack_steps_made_unviable = []
-        enabled_defenses = []
+        enabled_defenses = set()
+        attack_steps_made_unviable = set()
 
         for node in nodes:
             assert node == self.attack_graph.nodes[node.id], (
@@ -411,31 +412,24 @@ class MalSimulator():
             if node.is_available_defense():
                 node.defense_status = 1.0
                 node.is_viable = False
-                attack_steps_made_unviable += \
+                attack_steps_made_unviable |= \
                     apriori.propagate_viability_from_unviable_node(node)
                 agent.reward -= node.extras.get("reward", 0)
-                enabled_defenses.append(node)
+                enabled_defenses.add(node)
                 logger.info(
                     'Defender agent "%s" enabled "%s"(%d).',
                     agent.name, node.full_name, node.id
                 )
 
-
         for node in enabled_defenses:
             # Remove enabled defenses from defender action surfaces
             for defender_agent in self._get_defender_agents():
-                try:
-                    defender_agent.action_surface.remove(node)
-                except ValueError:
-                    pass
+                defender_agent.action_surface.discard(node)
 
         for attack_step in attack_steps_made_unviable:
             # Remove unviable attack steps from attacker action surfaces
             for attacker_agent in self._get_attacker_agents():
-                try:
-                    attacker_agent.action_surface.remove(attack_step)
-                except ValueError:
-                    pass
+                attacker_agent.action_surface.discard(attack_step)
 
         if self.sim_settings.uncompromise_untraversable_steps:
             self._disable_attack_steps(attack_steps_made_unviable)
@@ -443,8 +437,9 @@ class MalSimulator():
         return enabled_defenses, attack_steps_made_unviable
 
     def step(
-            self, actions: dict[str, list[AttackGraphNode]]
-        ) -> tuple[list[AttackGraphNode], list[AttackGraphNode]]:
+        # TODO: Do we care about order of requested actions to have a list?
+        self, actions: dict[str, list[AttackGraphNode]]
+    ) -> dict[str, MalSimAgentStateView]:
         """Take a step in the simulation
 
         Args:
@@ -452,35 +447,41 @@ class MalSimulator():
                   contains the actions for that user.
 
         Returns:
-        - (enabled_nodes, disabled_nodes)
+        - (compromised_nodes, new_attack_surface, disabled_nodes)
         """
-        logger.debug(
-            "Stepping through iteration %d/%d", self.cur_iter, self.max_iter)
+        logger.debug("Stepping through iteration %d/%d", self.cur_iter, self.max_iter)
         logger.debug("Performing actions: %s", actions)
 
-        enabled_nodes = []
-        disabled_nodes = []
+        disabled_nodes = set()
+        all_performed = set()
+
         all_attackers_terminated = True
 
-        # Peform agent actions
+        # Perform agent actions
         # Note: by design, defenders perform actions
         # before attackers (see _register_agent)
-        for agent_name in self.agents:
-            agent = self._agents_dict[agent_name]
-            agent_actions = actions.get(agent_name, [])
-            match agent.type:
+        for agent in self.alive_agents.values():
+            agent_actions = actions.get(agent.name, [])
 
+            agent.step_disabled_nodes = disabled_nodes
+
+            match agent.type:
                 case AgentType.ATTACKER:
-                    enabled = self._attacker_step(agent, agent_actions)
-                    enabled_nodes += enabled
+                    performed_steps, new_surface = self._attacker_step(
+                        agent, agent_actions
+                    )
+                    agent.step_action_surface = new_surface
+                    agent.step_compromised_nodes = performed_steps
+                    all_performed |= performed_steps
                     if not agent.terminated:
                         all_attackers_terminated = False
 
                 case AgentType.DEFENDER:
-                    enabled, disabled = \
+                    enabled_defenses, disabled = \
                         self._defender_step(agent, agent_actions)
-                    enabled_nodes += enabled
-                    disabled_nodes += disabled
+                    agent.step_compromised_nodes = all_performed
+                    agent.step_enabled_defenses = enabled_defenses
+                    disabled_nodes |= disabled
 
                 case _:
                     logger.error(
@@ -489,29 +490,25 @@ class MalSimulator():
                     )
 
         if all_attackers_terminated:
-            # Terminate all agents if all attackers are terminated
+            # Terminate all defenders if all attackers are terminated
             logger.info("All attackers are terminated")
-            for agent in self._agents_dict.values():
+            for agent in self.agents.values():
                 agent.terminated = True
 
         if self.cur_iter >= self.max_iter:
             # Truncate all agents when max iter is reached
             logger.info("Max iteration reached - all agents truncated")
-            for agent in self._agents_dict.values():
+            for agent in self.agents.values():
                 agent.truncated = True
 
-        agents_to_remove = set()
-        for agent_name in self.agents:
-            agent = self._agents_dict[agent_name]
+        for agent in self.alive_agents.copy().values():
             if agent.terminated or agent.truncated:
                 logger.info("Removing agent %s", agent.name)
-                agents_to_remove.add(agent.name)
-
-        for agent in agents_to_remove:
-            self.agents.remove(agent)
+                del self.alive_agents[agent.name]
 
         self.cur_iter += 1
-        return enabled_nodes, disabled_nodes
+
+        return self.agent_states
 
     def render(self):
         """Render attack graph from simulation in Neo4J"""
--- a/malsim/scenario.py
+++ b/malsim/scenario.py
@@ -137,10 +137,9 @@ def _validate_scenario_property_rules(
                 "observability/actionability rules"
             )
 
-    asset_names = set(a.name for a in graph.model.assets.values())
     for asset_name in rules.get('by_asset_name', []):
         # Make sure each specified asset exist
-        assert asset_name in asset_names, (
+        assert asset_name in graph.model._name_to_asset, (
             f"Failed to find asset name '{asset_name}' in model "
             f"'{graph.model.name}' when applying scenario" 
             "observability/actionability rules")
@@ -201,7 +200,7 @@ def apply_scenario_node_property_rules(
 
 def apply_attacker_entrypoints(
         attack_graph: AttackGraph, attacker_name: str, entry_points: dict
-) -> None:
+) -> int:
     """Apply attacker entrypoints to attackgraph from scenario
 
     Creater attacker, add entrypoints to it and compromise them.
@@ -234,7 +233,7 @@ def apply_attacker_entrypoints(
 
 
 def load_simulator_agents(
-        attack_graph: AttackGraph, scenario: dict, agent_config={}
+        attack_graph: AttackGraph, scenario: dict
     ) -> list[dict[str, Any]]:
     """Load agents to be registered in MALSimulator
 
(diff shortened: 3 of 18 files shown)


.github/workflows/test-pytest.yml
```diff
@@ -17,8 +17,8 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install .
+          pip install ".[ml]"
```
mrkickling, line 20: What does [ml] mean?

malsim/scenario.py
```diff
@@ -233,7 +233,7 @@ def apply_attacker_entrypoints(
 
 
 def load_simulator_agents(
-        attack_graph: AttackGraph, scenario: dict, agent_config={}
```
mrkickling, line 236: I recently added the agent_config as a kwarg to be able to set it for agents when loading agents from the scenario. Did you feel like it did not make sense?

malsim/sims/mal_sim_agent_state.py
```diff
@@ -56,32 +56,11 @@ class MalSimAgentStateView:
     def __init__(self, agent: MalSimAgentState):
         self._agent = agent
 
-    @property
-    def name(self) -> str:
-        return self._agent.name
+    def __post_init__(self) -> None:
```
mrkickling, line 59: I am not a fan of this. It is less lines of code, but otherwise "explicit is better than implicit". Overriding `__post_init` and `_freeze` seems unconventional to me as well.

When using a MalSimAgentStateView a linter wont be able to infer the instance variables of it now which makes using the MalSimAgentStateView cumbersome for someone used to autocompletion.

There might be some middle ground here that maybe inherits MalSimAgentState, but otherwise I prefer the old version.

malsim/sims/mal_simulator.py
```diff
@@ -389,27 +378,22 @@ def step(
...
                 logger.info("Removing agent %s", agent.name)
-                agents_to_remove.add(agent.name)
-
-        for agent in agents_to_remove:
-            self.agents.remove(agent)
+                del self.alive_agents[agent.name]
 
         self.cur_iter += 1
-        return enabled_nodes, disabled_nodes
+        return actioned_nodes, new_attack_surface, disabled_nodes
```
mrkickling, line 396: I think we discussed about returning AgentStateViews instead of dicts and include 'performed_actions' in AgentState. Wouldn't that be nicer, so we only return one list/dict?

malsim/sims/mal_sim_agent_state.py
mrkickling: Yeah, I read your initial description on the PR now :)

mrkickling:
> I implemented here most of my suggestions in #87. See the commit messages for what they do. They tests fail because the mal-toolbox version used here is missing [mal-lang/mal-toolbox#121](https://github.com/mal-lang/mal-toolbox/pull/121). When this gets merged, the tests will also pass. The suggestions I did not implement have to do with renaming folders/files as that would make following the changes harder. There are also a couple TODOs added imin the code about things not clear to me. Depending on what the answer is to those TODOs the code may be possible to simplify in some parts.

Makes sense, but we should first merge 121 then in my opinion. I will review that one now as well.
 
> While the above works, I think it would be better if we extended the `MalSimAgentState` class with three new keys to hold the three bits of information above:
 
>     # Contains possible actions for the agent in the next step
>     action_surface: set[AttackGraphNode] = field(default_factory=set)
>     new_action_surface: ....

Why `new_action_surface`? The `action_surface` is supposed to always be up to date and should be enough. 

>     actioned_nodes: ....

Yes, agreed, but `enabled_nodes` maybe?

>     disabled_nodes: ....

It is not clear to me what nodes this will contain for each agent? Maybe disabled nodes is not necessary? 

> If the above happens (trivial), then the return value of `step()` could change to sth simpler and more intuititve, i.e. a dict like:
> 
> ```
> {
>     "attacker1": MalSimAgentState(....for-attacker1...),
>     "defender1": MalSimAgentState(....for-defender1....),
> ```

I like this a lot more and I think we should implement this right now instead of postponing it.

mrkickling:
In general I like the PR, but I don't like the new implementation of `MalSimAgentStateView` as commented. I think we should implement the new return values of step() in this PR before merging (this includes adding `performed_nodes` to `MalSimAgentState(View)`. Otherwise minor comments and I wants the tests to pass before merging to new design branch, so I will take a look at the mal-toolbox PR you mentioned now.

.github/workflows/test-pytest.yml
nkakouros: Machine Learning. It could be RL, reinforcement learning, but once I did ML using the simulator.

malsim/scenario.py
nkakouros: I think it is a good idea. I just change it so that the config comes from the scenario file directly, without needing to do anything more than just adding it to the scenario file. Or I misunderstood sth? The function that would be called is `create_simulator_from_scenario` and in there the agent_config was expected to come as an arg but that put the burden on the user to somehow come up with the agent config programmatically, even if it were just pulling the config from the scenario file. What I suggest in this PR is to always look for the agent config in `load_simulator_agents`.

malsim/sims/mal_sim_agent_state.py
nkakouros: The reason I changed it is because:

- If we update the state class, we don't have to remember to update the view class.
- If we update the state class and need to update the view class as well, it is more probable to get bugs.
- The change I propose is pythonic and uses standard Python interfaces in the way they are supposed to be used. There is nothing hacky about this. So, the implementation is "explicit enough" without the pitfalls of copy-paste explicitness. 
- Comprehensibility is better I think. You read `__getattr__` and realize that it delegates to the parent class. You read `__setattr__` and see it's locked down. You don't have to search for each property, see if one is missing, where, why, etc.
- It reduces lines of code.
- It does not repeat the code, making it more elegant.

malsim/sims/mal_simulator.py
nkakouros: Yes, I agree.

nkakouros:
> > I implemented here most of my suggestions in #87. See the commit messages for what they do. They tests fail because the mal-toolbox version used here is missing [mal-lang/mal-toolbox#121](https://github.com/mal-lang/mal-toolbox/pull/121). When this gets merged, the tests will also pass. The suggestions I did not implement have to do with renaming folders/files as that would make following the changes harder. There are also a couple TODOs added imin the code about things not clear to me. Depending on what the answer is to those TODOs the code may be possible to simplify in some parts.
> 
> Makes sense, but we should first merge 121 then in my opinion. I will review that one now as well.

👍 

> > While the above works, I think it would be better if we extended the `MalSimAgentState` class with three new keys to hold the three bits of information above:
> 
> > ```
> > # Contains possible actions for the agent in the next step
> > action_surface: set[AttackGraphNode] = field(default_factory=set)
> > new_action_surface: ....
> > ```
> 
> Why `new_action_surface`? The `action_surface` is supposed to always be up to date and should be enough.

The action surface contains the history as well. It has the issue of growing linearly as the simulation proceeds. This variable here will contain only the newly available actions, the viable children of the just compromised nodes.

> > ```
> > actioned_nodes: ....
> > ```
> 
> Yes, agreed, but `enabled_nodes` maybe?

`enabled_nodes` works for defenders, but what does it mean for an attacker? It is confusing. I think the first reading is that this will contain the steps that became enabled due to the just compromised step(s). I thought of `performed_nodes` but you perform an action not a node. `performed_actions` misses the fact that these are `AttackGraphNode`s. I went with `actioned_nodes` because it captures both attackers and defenders and also nodes.

> > ```
> > disabled_nodes: ....
> > ```
> 
> It is not clear to me what nodes this will contain for each agent? Maybe disabled nodes is not necessary?

This will contain all disabled nodes due to defenders' actions. All agents will get the same copy (actually a pointer to the same data, so no memory overhead). It is needed to optimize things like the agent looking in its existing list of available next targets and removing those that got disabled. If this is not supplied to them, we end up in the situation we are now in where the agents needs to loop and search. This disabled_nodes is a set and it will be really fast to apply set operations to remove things from the agent's next_targets set.

> > If the above happens (trivial), then the return value of `step()` could change to sth simpler and more intuititve, i.e. a dict like:
> > ```
> > {
> >     "attacker1": MalSimAgentState(....for-attacker1...),
> >     "defender1": MalSimAgentState(....for-defender1....),
> > ```
> 
> I like this a lot more and I think we should implement this right now instead of postponing it.

Yes, I agree. I will provide an implentation soon (today).

.github/workflows/test-pytest.yml
mrkickling: But what does it do?

malsim/scenario.py
mrkickling: The idea was to have to do it programatically. But I think having it be defined in scenario is not a bad thing, so that is good.

.github/workflows/test-pytest.yml
nkakouros: It installs the `ml` extra dependencies defined in pyproject.toml. I added a `dev` extra package of dependencies as well.

malsim/sims/mal_sim_agent_state.py
mrkickling: - If we add a field to the state class we want to expose to read only, we update the view class. I agree that can be a bit annoying, but not a big issue in my opinion.

- The change might be pythonic in some sense, but it still makes it difficult to use with auto completion/type checking.

I also realized that the `MalSimAgentStateView` is not actually read-only in your solution. This is because `__post_init__` is not called automatically unless it is a dataclass. We should have a test to make sure the MalSimAgentView is read-only.

mrkickling:
> The action surface contains the history as well. It has the issue of growing linearly as the simulation proceeds. This variable here will contain only the newly available actions, the viable children of the just compromised nodes.

So the action_surface contains both performed and possible next step nodes while new_action_surface contains only possible next step nodes? Maybe just renaming `new_action_surface` to `non_performed_action_surface` would be clearer? 

> enabled_nodes works for defenders, but what does it mean for an attacker? It is confusing. 

I see. Why not performed_nodes?

> This will contain all disabled nodes due to defenders' actions. All agents will get the same copy (actually a pointer to the same data, so no memory overhead). It is needed to optimize things like the agent looking in its existing list of available next targets and removing those that got disabled. If this is not supplied to them, we end up in the situation we are now in where the agents needs to loop and search. This disabled_nodes is a set and it will be really fast to apply set operations to remove things from the agent's next_targets set.

But shouldn't that (the disabling) already have affected the agents action surface (which is supposed to be a source of truth for what an agent can do)? If they are now unviable the attacker should not have them in the action surface anymore?

I might be missing something though.

malsim/sims/mal_sim_agent_state.py
nkakouros: You are right, I 've been looking too long at my own code to realize what I'm doing. Check out the latest version of the view class. It now supports IDE completion. And REPL completion. And freezing agent properties to avoid mutating them. And at the same number of lines of code as the original version despite including doctstrings and being dynamic. What else can you ask for? 💪💪💪💪💪

nkakouros:
> > The action surface contains the history as well. It has the issue of growing linearly as the simulation proceeds. This variable here will contain only the newly available actions, the viable children of the just compromised nodes.
> 
> So the action_surface contains both performed and possible next step nodes while new_action_surface contains only possible next step nodes? Maybe just renaming `new_action_surface` to `non_performed_action_surface` would be clearer?

Not really. The new_action_surface contains only the new steps that became available due to the agent comprising the steps it did in the last time step. If you do password.brute_force for example new_action_surface will contain only service.login. But not VM.reboot which may have become available in the past.

> 
> > enabled_nodes works for defenders, but what does it mean for an attacker? It is confusing.
> 
> I see. Why not performed_nodes?

I quote:

> I thought of performed_nodes but you perform an action not a node. performed_actions misses the fact that these are AttackGraphNodes.

But if you feel strongly about it, I'm ok.

> > This will contain all disabled nodes due to defenders' actions. All agents will get the same copy (actually a pointer to the same data, so no memory overhead). It is needed to optimize things like the agent looking in its existing list of available next targets and removing those that got disabled. If this is not supplied to them, we end up in the situation we are now in where the agents needs to loop and search. This disabled_nodes is a set and it will be really fast to apply set operations to remove things from the agent's next_targets set.
> 
> But shouldn't that (the disabling) already have affected the agents action surface (which is supposed to be a source of truth for what an agent can do)? If they are now unviable the attacker should not have them in the action surface anymore?
> 
> I might be missing something though.

Yes, it will already have affected their action surface but not their internal variables that they may use to keep track of their own lists, like the BFS/DFS do where they keep a list of next actions ordered according to their policy. It is the agents' job to remove a disabled node from those lists. Giving them the set of disabled nodes ready, speeds this process up for them, as the don't have to loop these variable and check each one node if it is disabled or not. They can now just remove these nodes from their lists directly.

malsim/sims/mal_sim_agent_state.py
mrkickling: Haha, then I think I am happy now!

malsim/sims/mal_simulator.py
```diff
@@ -322,63 +308,64 @@ def _defender_step(
...
 
                 case AgentType.DEFENDER:
-                    enabled, disabled = \
+                    enabled_defenses, disabled = \
                         self._defender_step(agent, agent_actions)
-                    enabled_nodes += enabled
-                    disabled_nodes += disabled
+                    agent.step_compromised_nodes = all_performed
+                    agent.step_enabled_defenses = enabled_defenses
+                    disabled_nodes |= disabled
```
nkakouros, line 369: Defenders are never terminated. What happens if a defender has no more defenses to enable? Is it allowed to disable defenses?

mrkickling: Are defenders never terminated? I am not sure if I follow.

malsim/sims/mal_simulator.py
```diff
@@ -322,63 +307,67 @@ def _defender_step(
...
                     if not agent.terminated:
                         all_attackers_terminated = False
 
                 case AgentType.DEFENDER:
                     enabled, disabled = \
                         self._defender_step(agent, agent_actions)
-                    enabled_nodes += enabled
-                    disabled_nodes += disabled
+                    # TODO: do we expect the defenders to have a view of all 
+                    # actioned nodes, not all theirs?
```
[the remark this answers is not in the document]
mrkickling: Sounds good to me

malsim/sims/mal_simulator.py
nkakouros: If I understand correctly, attackers may get terminated when they execute their actions at a time step. Then, if all attackers are terminated, the defender gets terminated as well. Is the idea to always have a defender alongside attackers? Can't the defender be terminated on its own while attackers continue working? What happens if a defender runs out of defenses to enable?

malsim/sims/mal_simulator.py
nkakouros: Done.

malsim/sims/mal_simulator.py
mrkickling: This is probably something to bring into the mal simulator meetings. I tried to just copy the behavior from the old simulator. I think your understanding is correct:
- Attackers are terminated if they have nothing left to do
- Defenders are terminated if all attackers have nothing left to do

I guess if defenders running out of things to do would also mean that the attackers run out of things to do (since the defender has shut down the whole network), so in practice it works both ways. Do you agree?

kasanari: The defender never runs out of things to do since it can choose to wait.

mrkickling: Yes that is one way to see it, but nothing stops the attacker from waiting either I believe.

Saying 'nothing left to do' might have been misleading, what I meant is having nothing left in its action surface / having its whole action surface already be enabled.

kasanari: That attacker can not wait because the document says so /s. I agree that the situation where the defender action surface is empty could be handled in a different manner. "Fast-forwarding" the simulator to the end and aggregating the rewards would be one option.

nkakouros: A related question is if agents should be attached at the start of the simulation always, or if there should be rules (even if just time-based) for when they attach to the attack graph. But I go the answer I wanted. Feel free to resolve this if not needed anymore.

malsim/scenario.py
```diff
@@ -137,10 +137,9 @@ def _validate_scenario_property_rules(
                 "observability/actionability rules"
             )
 
-    asset_names = set(a.name for a in graph.model.assets.values())
     for asset_name in rules.get('by_asset_name', []):
         # Make sure each specified asset exist
-        assert asset_name in asset_names, (
+        assert asset_name in graph.model._name_to_asset, (
```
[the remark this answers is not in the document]
nkakouros: Sure, let's change back. I had it like this and got some conflict, and reverted to my change. This will be removed when the LookupDict will be merged and the model's `_name_to_asset` is removed. Then, I will remember  to revisit these lines of code with a grep. If you have it with `asset_names` it may still work but maybe doing an unneeded list comprehension and I doubt I will remember to change it.

malsim/mal_simulator.py
```diff
@@ -349,19 +345,23 @@ def _attacker_step(
...
-                            node.full_name)
+                               node.full_name)
+
+        # Update attacker action surface
+        new_attack_surface = query.calculate_attack_surface(
+            attacker, from_nodes=compromised_nodes, skip_compromised=True
+        )
+
+        # TODO: why extend?
+        agent.action_surface |= new_attack_surface
```
andrewbwm, line 364: What do you mean by why extend? This is just determining the attack surface from the node it just compromised. The old nodes in the attack surface are still potential options?

nkakouros: Right, it's a leftover reminder to discuss this. I will remove it.

nkakouros: Actually, the discussion is still to be done right? If the simulator should include the already compromised nodes in the attack surface sent to the attacker or not, which is sth the attacker knows already. If the surface is huge and the attacker needs to extract only the non-compromised ones, it makes it slow or at least more complicated than it needs to be. Perhaps the state could include the non-compromised nodes in the attack surface already?

malsim/scenario.py
```diff
@@ -201,7 +200,7 @@ def apply_scenario_node_property_rules(
 
 def apply_attacker_entrypoints(
         attack_graph: AttackGraph, attacker_name: str, entry_points: dict
-) -> None:
+) -> int:
     """Apply attacker entrypoints to attackgraph from scenario
 
     Creater attacker, add entrypoints to it and compromise them.
```
andrewbwm, line 206: The method should be renamed and the docstrings should be updated. I also think that it might be a reasonable idea to just return the Attacker that was created instead. But I really like the idea behind the change.

nkakouros: I thikn this is a comment for Joakim's PR. I just changed the return type annotation.

malsim/mal_simulator.py
andrewbwm: I think we might be combining two related discussions into one. This does not explicitly have anything to do with compromised nodes . It just means that the old values in the attack surface should still be part of it, whether some of them have already been compromised or not is irrelevant here.

I do remember/understand the point you were making that the attack surface scales with the size of the attack graph, which it absolutely does, as do the `reached_attack_steps` on the `AttackGraph` `Attacker`. And that we may wish to have optimized versions that just provide deltas and let the agents handle them as they see fit.

We may want this configurable. But there's no doubt that right now it should keep the previous entries in the attack surface, it replaces the old update method.

malsim/mal_simulator.py
```diff
@@ -411,76 +412,76 @@ def _defender_step(
...
 
         if self.sim_settings.uncompromise_untraversable_steps:
             self._disable_attack_steps(attack_steps_made_unviable)
 
         return enabled_defenses, attack_steps_made_unviable
 
     def step(
-            self, actions: dict[str, list[AttackGraphNode]]
-        ) -> tuple[list[AttackGraphNode], list[AttackGraphNode]]:
+        # TODO: Do we care about order of requested actions to have a list?
```
andrewbwm, line 440: Yes, we do, if we allow an agent to take multiple actions at once they need to happen in order in case some of  the actions earlier in the order make possible actions later in the list.

nkakouros: This may complicate things because the input to the step methods needs to be a list then, and these methods will return a set. Then, the agent needs to consume a set and return a list. Maybe it is simple to change that but I'll have to fix the code.

malsim/scenario.py
nkakouros: Right! But what should this be renamed to? What is wrong here?

andrewbwm: `Apply` to me suggestions a non-return methods. Now, don't get me wrong, `add` also does that and we have return values on plenty of `add` methods, but I prefer it.

malsim/scenario.py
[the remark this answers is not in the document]
nkakouros: Done.

malsim/mal_simulator.py
nkakouros: I tried playing with the code a bit, and returning all possible actions in the attack surface makes sense, even compromised ones. The operation to extract the uncompromised steps from the attack surface will be the same regardless of whether it happens in the agent or the simulator. Since some agents may not need this extraction, it's better left to the agents' side to occur on demand.

malsim/scenario.py
nkakouros: I tried renaming this and have it return an Attacker and then got into the rabbit hole of using `attacker_id` instead of `Attacker` objects and trying to fix that and the code became so convoluted that it just did not make sense to me. I will leave this to you or Joakim to fix as you please. My suspicion is that this will be greatly simplified when the Attacker class is removed (or whatever it ends up being with node.compromise/attacker.compromise).

malsim/mal_simulator.py
nkakouros: Actually, I did not change the signatures of the args in the step methods :phew:, they are still lists.

nkakouros:
If you merge https://github.com/mal-lang/mal-toolbox/pull/123, this will pass. I don't think I have anything more to do here.

nkakouros:
Additionally, I renamed wrappers to envs and fixed a bug where numpy was needed to import anything despite it being an optional dependency now.

nkakouros:
I also added a stricter check when accepting to compromise a node in _attacker_step as per [this comment](https://github.com/mal-lang/mal-simulator/pull/87#discussion_r1959985952). This makes two tests fail and I think they might be buggy.

malsim/mal_simulator.py
```diff
@@ -33,66 +38,81 @@ class MalSimAgentState:
...
         super().__init__(name, AgentType.DEFENDER)
 
 
-class MalSimAgentStateView:
+# Generic T is used here to allow IDEs to provide autocompletions from
+# MalSimAgentState.
+T = TypeVar("T", bound=MalSimAgentState)
+
+
+class MalSimAgentStateView(Generic[T]):
```
mrkickling, line 81: What does the Generic[T] give us?

malsim/mal_simulator.py
```diff
@@ -33,66 +38,81 @@ class MalSimAgentState:
     # Attackers get positive rewards, defenders negative
     reward: int = 0
 
-    # Contains possible actions for the agent in the next step
-    action_surface: list[AttackGraphNode] = field(default_factory=list)
+    # Contains possible actions for the agent on the next step
+    action_surface: set[AttackGraphNode] = field(default_factory=set)
+
+    # Contains possible actions that became available on the last step
+    step_action_surface: set[AttackGraphNode] = field(default_factory=set)
```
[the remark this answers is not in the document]
nkakouros: I don't like it much either but I couldn't find sth better and shorter. I also think it is good to have a common naming scheme for all 3 or so properties that refer to the new additions, etc. A name I tried and kind of liked was just_performed_steps, etc. But that didn't go well with just_action_surface. Feel free to rename these if you come up with sth better

malsim/mal_simulator.py
nkakouros: In this case, it is a way to have static analysis pick the properties of the agent state class as valid for the view object too.

malsim/mal_simulator.py
```diff
@@ -33,66 +38,81 @@ class MalSimAgentState:
...
-        return self._agent.terminated
+    def __getattr__(self, attr) -> Any:
+        """Return read-only version of proxied agent's properties."""
+        value = getattr(self._agent, attr)
 
-    @property
-    def action_surface(self) -> list[AttackGraphNode]:
-        return self._agent.action_surface
+        if isinstance(value, dict):
+            return MappingProxyType(value)
```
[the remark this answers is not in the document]
nkakouros: Yes, everything.

malsim/mal_simulator.py
```diff
@@ -268,31 +274,22 @@ def register_defender(self, name: str):
...
-        """Return read only agent state for agent with given name"""
-
-        assert name in self._agents_dict, (
-            f"Agent with name '{name}' does not exist")
-        agent = self._agents_dict[name]
-        return MalSimAgentStateView(agent)
-
-    def get_agents(self) -> list[MalSimAgentStateView]:
+    @property
+    def agent_states(self) -> list[MalSimAgentStateView]:
```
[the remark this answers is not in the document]
mrkickling: Also, I will repeat that we should not replace all `self.get_x` with `self.x.get(y, default)`. As we can see in this example it removes an error check (and it hides the api). But we will not discuss it more now but instead at some point do an overview of the whole project after we have discussed it together and decided on one approach. Making an issue for it.

malsim/mal_simulator.py
andrewbwm: I'm also not a fan of code that is in place only to facilitate specific static analyses. On a broader point we're running into our old problem again. I will be merging a version of this in tomorrow, the more "extra" stuff like this you add the more likely I am to just merge a minimal version instead of most of it.

So, it's your choice how you wish to approach this. We have to move forward, and I'm not opposed to you having a forked version of this. But if you decide that you'd rather we keep a common development trunk this kind of addition is not helpful.

That isn't to say it isn't reasonable for you to add this type of code, it's just clearly beyond the scope of the PR this is supposed to be merged into. A pull request which is plenty complex as is anyway.

malsim/mal_simulator.py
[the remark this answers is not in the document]
mrkickling: Either way this works fine for now, so feel free to resolve this.

malsim/mal_simulator.py
```diff
@@ -33,66 +38,81 @@ class MalSimAgentState:
...
     def __init__(self, name: str, attacker_id: int):
         super().__init__(name, AgentType.ATTACKER)
         self.attacker_id = attacker_id
 
 
 class MalSimDefenderState(MalSimAgentState):
     """Stores the state of a defender in the simulator"""
+
+    # Contains defense steps successfully enabled by the defender in the last step
+    step_enabled_defenses: set[AttackGraphNode] = field(default_factory=set)
```
mrkickling, line 70: I am sceptical to having two different variable names for 'latest enabled nodes' between attackers and defenders. Couldn't they both have `enabled_nodes` instead? I have a feeling we talked about this and you called them `actioned_nodes`. But I never understood that they would be different for Attackers/Defenders states.

malsim/mal_simulator.py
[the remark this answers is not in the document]
nkakouros: Yes it is a dict.

malsim/mal_simulator.py
nkakouros: The use of Generic[T] is the correct solution to ensure static analysis and IDE completion work with the dynamic view class, directly addressing Joakim's review comments. His point was valid for writing more comprehensive code, and this change enhances the developer experience while staying within the scope of the target PR.

I don't suggest changes for the sake of having discussions. What I propose addresses real issues with the code, they are not personal preferences. You can dismiss them but that will not improve the code.

The implication that I'm delaying progress feels unfair, especially when this specific change directly addressed prior feedback. I also took on the burden of implementing both my suggestions and some of yours in the target PR, trying to move things along when, after two weeks of little activity in that PR, I felt pressed by my own project deadlines. I get that reviewing shouldn't be a full-time job, but I broke the PR into bite-sized commits to make it easier. I only squashed them later because file renames in the target PR didn't wait for mine to get merged and left no other choice.

I completely understand the need to move forward. But prioritizing quick merges over quality is shortsighted and risks more technical debt—like the issues we’ve already tackled in the lang graph, the model redesign, and the simulator overhaul. I'm not the project maintainer or manager, but as a stakeholder relying on this project and with a need to extend it and make it run faster, I have a legitimate interest in its quality and sustainability. Maybe everyone keeping up with conflicts and feature branch maintenance could help us move development and reviews forward in parallel. There's no reason it has to be either development or reviews—both can progress together. Or perhaps we should discuss ways to improve the speed of reviews on both sides. They should not take but a small fraction of our work hours to process.

malsim/mal_simulator.py
nkakouros: To me enabled attack steps means steps that were enabled for the attacker to perform, not that the attack just performed. If you believe it's just me that could fall for this, feel free to change it back.

malsim/mal_simulator.py
nkakouros, line 278:
```suggestion
    def agent_states(self) -> dict[str, MalSimAgentStateView]:
```

malsim/mal_simulator.py
```diff
@@ -411,107 +414,102 @@ def _defender_step(
...
                     if not agent.terminated:
                         all_attackers_terminated = False
 
                 case AgentType.DEFENDER:
-                    enabled, disabled = \
+                    enabled_defenses, disabled = \
                         self._defender_step(agent, agent_actions)
-                    enabled_nodes += enabled
-                    disabled_nodes += disabled
+                    agent.step_compromised_nodes = all_performed
```
nkakouros, line 483: I now see that there is `MalSimulatorSettings.cumulative_defender_obs` that is not used anywhere at the moment. Do we want to respect that here or remove it from the settings completely?

malsim/mal_simulator.py
```diff
@@ -33,66 +38,81 @@ class MalSimAgentState:
...
     """Stores the state of a defender in the simulator"""
+
+    # Contains defense steps successfully enabled by the defender in the last step
+    step_enabled_defenses: set[AttackGraphNode] = field(default_factory=set)
+
     def __init__(self, name: str):
         super().__init__(name, AgentType.DEFENDER)
 
 
-class MalSimAgentStateView:
```
mrkickling, line 62: I noticed that autocompletion (in VS Code) still can not understand what instance variables an object of MalSimAgentStateView has. It makes it difficult for me to work with the MalSimAgentStateView objects, since I every time have to go to MalSimAttacker/DefenderState to find the available variables. I don't know why that is since I see you implemented dir which I thought would fix this (do I just have a bad autocompletion in my IDE?). 

Do you have autocompletion that works with this approach?

malsim/mal_simulator.py
mrkickling: Yes, I also thought about that. It was used in the serialized obs env before, but not anymore. I think it should be removed for less confusion, the 'observation' concept is not really clear in the MalSimulator so it is just confusing. Feel free to remove it. Of course this leads to us having just one setting in MalSimSettings, but that is okay for now.

malsim/mal_simulator.py
mrkickling: It seems like `__dir__` is not evaluated when doing static type checking, which makes sense I guess (return value from `__dir__` could be different in runtime so it can not be statically analyzed). 

Either:

1. Remove the things added for static type checking since it does not work (at least in pyright) and we live without those features for MalSimAgentStateView.

or

2. Go back to the @property solution for each instance variable.

or

3. Find another solution where auto completion works

I would say option 2 is simplest but I know you disliked it for other reasons.

malsim/mal_simulator.py
mrkickling: >To me enabled attack steps means steps that were enabled for the attacker to perform, not that the attacker just > performed. If you believe it's just me that could fall for this, feel free to change it back.

Ah I see, to me it means that the node was activated by an agent and I never saw the confusion but that explains why you would call them `actioned_nodes` instead and that is fine to me as well.

But I need to wrap my head around all the structures.

AgentState contains:
- `action_surface` - possible nodes to pick from next step
- `step_action_surface` - nodes that were added to action surface previous step 
- `step_compromised_nodes` - nodes that were compromised in previous step (by all attacker agents)
- `step_disabled_nodes` - nodes made unviable in previous step (by all defender agents)

DefenderAgentState also has:
- `step_enabled_defenses` - defenses enabled by just this specific defender

I would like to see something like `step_enabled_defenses` for attackers as well so you can easily see what nodes an attacker agent compromised in the last step. 

We could of course have a separate `step_compromised_nodes` for AttackerAgentState but I would much rather see an attribute in MalSimAgentState for both attackers/defenders (replacing `step_enabled_defenses`), called `step_performed_nodes` or `step_actioned_nodes` which contains the nodes the agent (defender or attacker) activated in the last step.

Maybe this is how is was before, in that case I must have misunderstood it previously, sorry in that case.

malsim/mal_simulator.py
nkakouros: __dir__ is for REPLs like the cli you get running ipython, so that they can get the dynamic properties of objects for their completion. Inheriting from the generic type is what should make IDE completion work. It works for me using pylsp. How do you have your completion setup?

malsim/mal_simulator.py
```diff
@@ -148,28 +167,22 @@ def __init__(
...
-        self.model = attack_graph.model
 
         self.sim_settings = sim_settings
-        self.max_iter = max_iter # Max iterations before stopping simulation
+        self.max_iter = max_iter  # Max iterations before stopping simulation
         self.cur_iter = 0        # Keep track on current iteration
 
         # Keep track on all registered agent states
-        self._agents_dict: dict[str, MalSimAgentState] = {}
+        self.agents: dict[str, MalSimAgentState] = {}
```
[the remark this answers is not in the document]
mrkickling: I can make this change myself though, and can be done after the merge.

malsim/mal_simulator.py
nkakouros: I couldn't get Generic to work as I thought it did, I think my completion engine is buggy and gave me false positives. If we 100% need editor completion we can do two things:

- `class MalSimAgentStateview(MalSimAgentState)`. The view will inherit from state and completion will pick the parent's properties fine. Of course, this does not make sense business-wise, but it will satisfy static analysis. Since we override `__getattr__` I don't think we risk parent properties messing the view object; we have tight control over what properties the view will support and how (frozen). With a comment that explains the situation, it may be an acceptable solution.

- Explicitly list the state properties in a literal type:
```
state_props = typing.Literal["action_surface", "name", ....]
```

This is simpler and saner but it requires duplication of the prop names.

There is always the possibility to not support autocompletion in IDEs.

malsim/mal_simulator.py
nkakouros: I removed it.

malsim/mal_simulator.py
nkakouros: I removed Generic. If we want to add support for this we can do maybe in another PR?

malsim/mal_simulator.py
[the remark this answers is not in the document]
nkakouros: This makes sense. 👍

malsim/mal_simulator.py
```diff
@@ -268,31 +277,22 @@ def register_defender(self, name: str):
...
         """Return list of mutable defender agent states"""
-        return [a for a in self._agents_dict.values()
-                if a.type == AgentType.DEFENDER]
+        return [a for a in self.agents.values() if a.type == AgentType.DEFENDER]
 
     def _disable_attack_steps(
-            self, attack_steps_to_disable: list[AttackGraphNode]
-        ):
+        self, attack_steps_to_disable: set[AttackGraphNode]
+    ):
```
nkakouros, line 295: I can't a comment to the whole body of this method, but I think it needs some attention. The name seems a bit off, it does not disable steps. Since it is specific to attackers, I would say it uncompromises steps. 

Also, the docstring references `observed_state` which is removed now.

A question is also if this should work on dead agents as well. If an agent is dead and an attack step executed in the past now becomes uncompromised, it's not reasonable to penalize that agent (although it's dead, it doesn't care) or to give back to the attacker points for that.

malsim/mal_simulator.py
```diff
@@ -411,107 +416,101 @@ def _defender_step(
...
 
-        for attack_step in attack_steps_made_unviable:
-            # Remove unviable attack steps from attacker action surfaces
-            for attacker_agent in self._get_attacker_agents():
-                try:
-                    attacker_agent.action_surface.remove(attack_step)
-                except ValueError:
-                    pass
+        for attacker_agent in self._get_attacker_agents():
+            defender_agent.step_action_surface_removals |= (
```
nkakouros, line 436: `defender_agent` here is the last value that variable got in the loop above. Was this meant to be a nested for loop in the above loop?

If so, I think these should updated at the end of the `step` method when all removed nodes, etc are collected.

malsim/mal_simulator.py
```diff
@@ -411,107 +416,101 @@ def _defender_step(
...
+                attacker_agent.action_surface & attack_steps_made_unviable
+            )
+            attacker_agent.action_surface -= attack_steps_made_unviable
 
         if self.sim_settings.uncompromise_untraversable_steps:
             self._disable_attack_steps(attack_steps_made_unviable)
```
nkakouros, line 443: The first line uses the setting called `uncompromised untraversable steps`. The second uses as argument the variable `attack steps made unviable`. Untraversable and viable are related but not the same. Either the setting should be renamed or it meant to do sth different to what the second line does?

malsim/mal_simulator.py
```diff
@@ -33,66 +34,88 @@ class MalSimAgentState:
...
+        field(default_factory = set))
+
+    # Contains previously possible actions that became unavailable in the last
+    # step
+    step_action_surface_removals: set[AttackGraphNode] = (
+        field(default_factory = set))
 
-    # Fields that tell if agent is 'dead' / disabled
+    # Contains nodes that defender actions made unviable in the last step
+    step_unviable_nodes: set[AttackGraphNode] = field(default_factory=set)
```
nkakouros, line 53: This is the same as `step_action_surface_removals` minus the enabled defense steps, right? It is a global view. Do you think it belongs to the attackers' state?

nkakouros:
I added b720c7e as a suggestion for rewards to happen out of the `_attacker_step` and `_defender_step` methods. Feel free to revert if you don't like it as it is an optimization for when there are more than one defenders.

malsim/mal_simulator.py
```diff
@@ -148,28 +159,22 @@ def __init__(
...
     def reset(
-            self,
-            seed: Optional[int] = None,
-            options: Optional[dict] = None
-        ) -> dict[str, MalSimAgentState]:
+        self,
+        seed: Optional[int] = None,
+        options: Optional[dict] = None
+    ) -> dict[str, MalSimAgentStateView]:
         """Reset attack graph, iteration and reinitialize agents"""
```
nkakouros, line 179: Sth that we missed here, given the new state properties, we should re-initialize those as well. The easiest way would be to just set self.agent_states to new state objects that use the same name and attacker_id. But with the current code and method layout it requires a bit of work. I will leave this to you as I am unsure of what the best approach here is.

nkakouros:
I think I am done with this PR. I don't have any other comments to make that are related to introducing #87.

malsim/mal_simulator.py
mrkickling: Yes, this is a config setting we have not used a lot. I agree on the naming and the docstring, will update that.

Good point about the dead agent scenario, I will look at it but leave it as a TODO if it is not trivial, since it only is relevant in a case where that setting (uncompromise_untraversable_steps) is enabled and more than one attacker agent exists. Neither of these are a use case for us or anyone I can think of,

malsim/mal_simulator.py
mrkickling: This is a typo/bug, it should be attacker_agent.

malsim/mal_simulator.py
mrkickling: `uncompromise_untraversable_steps` is a good name of the setting to me. The function ` disable_attack_steps` will probably change as per a different comment you had. That function will gets attack steps made unviable in this case, but only the ones that are compromised will be uncompromised, which matches the name of the setting. 

All traversable attack steps are viable, steps that have been made unviable are no longer traversable. That explains to me why the function takes attack_steps_made_unviable and it makes sense to me.

malsim/mal_simulator.py
```diff
@@ -33,66 +34,88 @@ class MalSimAgentState:
...
+    action_surface: set[AttackGraphNode] = field(default_factory=set)
+
+    # Contains possible actions that became available in the last step
+    step_action_surface_additions: set[AttackGraphNode] = (
+        field(default_factory = set))
+
+    # Contains previously possible actions that became unavailable in the last
+    # step
+    step_action_surface_removals: set[AttackGraphNode] = (
+        field(default_factory = set))
```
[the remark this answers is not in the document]
mrkickling: Could it be that this understanding is based on the bug/typo you commented on in defender_step?

It is supposed to contain what is says:
- All nodes that were removed from the agents action surface this step

malsim/mal_simulator.py
mrkickling: I will look into it.

malsim/mal_simulator.py
```diff
@@ -243,20 +255,17 @@ def _register_agent(self, agent: MalSimAgentState):
...
-        if agent.type == AgentType.DEFENDER:
-            # Defender is first in list so it can pick
-            # actions before attacker when step performed
-            self.agents.insert(0, agent.name)
-            self.possible_agents.insert(0, agent.name)
-        elif agent.type == AgentType.ATTACKER:
-            # Attacker goes last
-            self.agents.append(agent.name)
-            self.possible_agents.append(agent.name)
+        self.agents[agent.name] = agent
```
mrkickling, line 247: Not sure if I commented this somewhere else, but MalSimulator.agents should be private, right? We don't want users to fetch the AgentStates directly but instead to fetch the AgentStateViews.

malsim/mal_simulator.py
```diff
@@ -268,31 +277,22 @@ def register_defender(self, name: str):
...
-        return MalSimAgentStateView(agent)
-
-    def get_agents(self) -> list[MalSimAgentStateView]:
+    @property
+    def agent_states(self) -> dict[str, MalSimAgentStateView]:
         """Return read only agent state for all dead and alive agents"""
-        return [self.get_agent_state(agent) for agent in self.possible_agents]
+        return {name: MalSimAgentStateView(agent) for name, agent in self.agents.items()}
 
     def _get_attacker_agents(self) -> list[MalSimAttackerState]:
```
mrkickling, line 271: `_get_attacker/defender_agents` used to only return 'living' agents

mrkickling: Might have been wrong about this actually. Nevermind.

malsim/mal_simulator.py
mrkickling: Made it private now.

malsim/mal_simulator.py
mrkickling: Fixed

malsim/mal_simulator.py
mrkickling: Fixed the name/docstring

malsim/mal_simulator.py
mrkickling: Yes, I will probably move back to something closer to the old solution where we keep track of dead / living agents using their names, and store the states separatly. It feels odd that we store the states in two different dicts currently. I will push a commit and you can see what you think.

This will also make reset simpler.

malsim/mal_simulator.py
nkakouros: OK, it makes sense now.

malsim/mal_simulator.py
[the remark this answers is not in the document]
nkakouros: > Could it be that this understanding is based on the bug/typo you commented on in defender_step?

Yes, it makes sense now.

malsim/mal_simulator.py
andrewbwm: Correct. Viability is a property of the node alone while traversability is dependent on the attacker. A node is traversable if it is viable and any/all(or/and) of its necessary parents have been compromised.

nkakouros: Great, thanks.

PR #99, opened 2025-02-18
Add mypy workflow

nkakouros:
With the changes happening in #87, types are a bit out of sync in places. This adds mypy in strict mode as a CI job. Ideally, this mypy workflow and the ruff one would exist only in the mal-toolbox repo and be used here as reusable workflows. This will give a common static analysis coverage on both projects (and maybe others too).

### DIFF
--- a/.github/workflows/mypy.yml
+++ b/.github/workflows/mypy.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install -r tests/requirements.txt
+          pip install .[all]
       - name: Type checking with MyPy
         run: |
           pip install mypy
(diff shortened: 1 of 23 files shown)


mrkickling:
I fixed most of the errors in malsim, but I realized I missed ./tests. This PR is work in progress still so I will put it into Draft.

mrkickling:
Now all CIs are passing.

I had to do a few things:
- Add typehints to everything in the code base (I decided to ignore the gym_envs which I don't understand well enough, believe me I tried)
- Ignore missing imports since maltoolbox does not provide a required file to show that is supports mypy (maybe you know better how that works)
- Change so the CI install scripts used .[ml] and .[dev] instead of .[all] (since that was not defined)

I think this is mergeable. It does give mypy support for most of the code base. We should create an issue in the maltoolbox repo so we don't have to use 'ignore missing imports'. We should also typehint (or deprecate) the gym_envs.

PR #104, opened 2025-02-26
Have keyboard agent (attacker) only show not yet compromised nodes when asking user to pick action

mrkickling:
We have a KeyboardAgent that lets the user pick actions.

If an attacker is run using the KeyboardAgent, the user will be allowed to pick the same attack step as many times as they want to since it does not disappear from an attacker action surface, unlike defenders.  With the current implementation the action surface only grows and therefore the list of possible attacker choices in the KeyboardAgent also grows until it contains all attack steps.

This is not useful for users that want to use the KeyboardAgent.

### DIFF
--- a/malsim/agents/keyboard_input.py
+++ b/malsim/agents/keyboard_input.py
@@ -33,23 +33,26 @@ class KeyboardAgent(DecisionAgent):
             except ValueError:
                 return False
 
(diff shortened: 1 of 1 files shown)


mrkickling:
Not needed anymore after #108 was merged

PR #105, opened 2025-02-26
Malsim agent state static type checking

mrkickling:
MalSimAgentStateView currently can not be statically type checked and can not support auto completion in an IDE / editor (correct me if I am wrong and it's just my type checker that is missing some feature).

It seems to be hard to support type checking with the current approach. 
Personally, I find it hard to work with auto completion and my built in type checking does not work with the AgentStateView objects (it works for every other object in the simulator).

Sadly the only approach I could find from some googling / gpting was to go back to the properties route.

I added two StateViews to differentiate between attackers and defenders.
I also added a test that ensures that the attributes of the state view matches the attributes of the state. This makes sure there is never a mismatch in what attributes are supported in the state vs the view.

Let me know about any input and how to possibly make it less verbose. I would much rather have the simpler approach but I really also want to have auto complete..

### DIFF
--- a/malsim/mal_simulator.py
+++ b/malsim/mal_simulator.py
@@ -4,8 +4,7 @@ import copy
 from dataclasses import dataclass, field
 import logging
 from enum import Enum
-from types import MappingProxyType
-from typing import Any, Optional
+from typing import Optional
 
 from maltoolbox import neo4j_configs
 from maltoolbox.ingestors import neo4j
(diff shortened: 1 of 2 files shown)


nkakouros:
The changes here seems to me fall into what Andrei [calls](https://github.com/mal-lang/mal-simulator/pull/98/files/cef8b556851b812252946b564c88377c77fd7000#r1968085328) "code that is in place only to facilitate specific static analyses".

There are multiple alternative solutions. One is to learn the properties. I understand it is hard.

Another solution is autocompletion that pulls words from the currently open buffer. Since state is in mal_simulator.py, its properties will popup when you work with view objects in that file. You type e.g. `state_agent.step_` and then you get the three words that match.

Another solution is to use sth like [this](https://marketplace.visualstudio.com/items?itemName=Atishay-Jain.All-Autocomplete) that pulls words from all open files, not just the current file. This way you will have support for these words in other files as well.

Yet another solution is to have `MalSimAgentStateView` inherit from `MalSimAgentState`. This is what I would do. It is not "correct" in terms of business logic, but it will satisfy your static analysis and it will not cause "business" issues because we override `__getattr__` anyway and state properties will not be accessible directly from the view. With a comment that explains the situation, it will work fine.

There is one more solution that is 100% correct and THE solutions but works only with VS Code (which uses pylance) and not with [mypy yet](https://github.com/python/mypy/issues/8203). So, I won't mention it at all. When mypy supports it, it could be used to replace inheriting from the state class.

mrkickling:
This is code to facilitate static type checking, to make it easier for users (including me) to work with the objects.
Auto completion / static type checking works for all classes in the whole project except for this one.
And this is a class that has many similarly named attributes that are not easy to remember.

> Another solution is autocompletion that pulls words from the currently open buffer. Since state is in mal_simulator.py, its properties will popup when you work with view objects in that file. You type e.g. state_agent.step_ and then you get the three words that match.

For me it does not work on AgentStateViews, if that worked I would be more happy.

> Another solution is to use sth like [this](https://marketplace.visualstudio.com/items?itemName=Atishay-Jain.All-Autocomplete) that pulls words from all open files, not just the current file. This way you will have support for these words in other files as well.

I tried this now, but this is not what I want. This addon works with variable names, not on attributes of objects.

> Yet another solution is to have MalSimAgentStateView inherit from MalSimAgentState. This is what I would do. It is not "correct" in terms of business logic, but it will satisfy your static analysis and it will not cause "business" issues because we override __getattr__ anyway and state properties will not be accessible directly from the view. With a comment that explains the situation, it will work fine.

This would be an improvement for me. I tried this approach in this PR but could not make it work, but I can try again.

I think the solution proposed in this PR and the solution where MalSimAgentStateView inherit from MalSimAgentState would easily work for all static type checkers and it would make it easier for people using auto completion/type checking.

Do you think the solution in this PR is too many lines of code or non-pythonic, or do you think it is not a problem that MalSimAgentStateView does not have type checking/autocompletion support? I think my view here has changed, before I would never have cared about typing this much in python, but I guess my programming environment has changed to be more dependent on it.

nkakouros:
If you just inherit for the state class, it works fine:

Screenshot 2025-02-28 at 01 55 34

The above is on neovim with pyslp. 

And here you can see vs code:

Screenshot 2025-02-28 at 01 56 43

mrkickling:
Yes that works for the LSP/type analysis, but it definitely breaks things. Not possible to run a simulation with those agents with the cli as well as some of the tests  (try running `malsim tests/testdata/bfs_vs_bfs_scenario.yml`. it will not terminate agents because the MalSimAgentStateView will not refer to the same MalSimAgentState as before).

nkakouros:
See #109. I added a couple of lines and comments to explain what is going on.

mrkickling:
Not relevant any more

PR #108, opened 2025-02-27
Remove compromised nodes from the attack surface for now

andrewbwm:
This should eventually be a configuration. Some simulations might want to be able to go through attack steps multiple times.

### DIFF
--- a/malsim/mal_simulator.py
+++ b/malsim/mal_simulator.py
@@ -211,7 +211,9 @@ class MalSimulator():
 
             if isinstance(agent, MalSimAttackerState):
                 attacker = self.attack_graph.attackers[agent.attacker_id]
-                agent.action_surface = query.calculate_attack_surface(attacker)
+                agent.action_surface = query.calculate_attack_surface(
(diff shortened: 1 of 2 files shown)


malsim/mal_simulator.py
```diff
@@ -350,11 +352,16 @@ def _attacker_step(
...
         attack_surface_additions = query.calculate_attack_surface(
-            attacker, from_nodes=compromised_nodes, skip_compromised=True
+            attacker, from_nodes = compromised_nodes, skip_compromised = True
         )
 
         # Filter out nodes already in action surface, these are not additions.
         attack_surface_additions -= agent.action_surface
+        # Also filter out nodes already compromised from the attack surface.
+        # TODO: Make this configurable in the future
+        attack_surface_additions -= compromised_nodes
```
mrkickling, line 362: Should this not already be filtered out since we called `query.calculate_attack_surface` with `skip_compromised` after the selected action nodes are already compromised, or is there some edge case I didn't think of?

PR #109, opened 2025-02-28
Support completions for dynamic View class

nkakouros:


mrkickling:
Awesome, and thank you for finding a solution that satisfies the static type checking. Do we still actually need `__dir__` or will it be inferred from the class inheritance now?

nkakouros:
> Do we still actually need `__dir__` or will it be inferred from the class inheritance now?

It is still needed. Attributes without primitive defaults (0, False) will not show up otherwise. 

I pushed a commit to properly handle dunder properties and methods as well. 

The View class is a generic class that is not specific to the agent state or the mal simulator at all. I would move it to a utils.py module and rename it to sth generic like ObjectView or Object Proxy.

andrewbwm:
> I did think about renaming to AgentState, AttackerState, Defenderstate. MalSim is a bit unnecessary and makes the class name very long.

I'm for it, I almost did that when I was going through the PRs.

Issue #101, opened 2025-02-24
Print step progression regularly


nkakouros:
Currently, there is no no output for the simulator with the default log level about the progression of the simulation. It would be nice to have such a info log that outpus sth like:

```
Timestep 2000/2341343 (0.3s)
```

where `2000` is the current step, `2341343` is the max_iter setting and `0.3s` is for how long the simulator has been running for. 

To make this useful, the stride step (how often in terms of timesteps this gets printed) should be dynamic and depend on max_iter.

I have an implementation for this, let me know if it sth you would like to have in the simulator.

mrkickling:
It could be useful. Jakob uses `tqdm` to see progress during simulations, could that be an alternative? Maybe it is not for logging to files but only for stdout, I am not sure. It would be interesting to see your implementation.

nkakouros:
This is how I calculate stride (from chatgpt):

```python
def calculate_stride_size(value: int) -> int:
    """
    Calculates frequency using order of magnitude and first digit of value.

    If value is too small (less than 150) sets the frequency to value, i.e.
    there is no repetition.

    Args:
        value: The total number of iterations.
    """
    if value <= 0:
        msg = "Value myst be a positive number"
        logger.error(msg)
        raise ValueError(msg)
    if value < 150:
        stride_size = value
    else:
        order_of_magnitude = math.floor(math.log10(value))
        first_digit = int(str(value)[0])

        base = 10 ** max(0, order_of_magnitude - 1)  # e.g., for 1000s, base is 100
        multiplier = max(1, first_digit)  # Multiplier based on first digit
        stride_size = base * multiplier
        stride_size = max(30, int(stride_size))  # Minimum frequency of 30

    return stride_size
```

This helps produce logs like this:

```
2025-03-03 14:10:26 - INFO [simulate:69]: Simulation started.                                      
2025-03-03 14:10:26 - INFO [simulate:80]: Timestep      0/173343. Time: 0.00s.                     
2025-03-03 14:10:26 - INFO [simulate:83]: Timestep      2/173343: Deploying bfs-random attacker.   
2025-03-03 14:10:37 - INFO [simulate:80]: Timestep  10000/173343. Time: 11.02s.                    
2025-03-03 14:11:09 - INFO [simulate:80]: Timestep  20000/173343. Time: 42.47s.                    
2025-03-03 14:11:38 - INFO [simulate:80]: Timestep  30000/173343. Time: 71.51s.                    
[...]

PR #110, opened 2025-02-28
Less verbose logging when used as lib

nkakouros:
When running the simulator as an app (`malsim` command), debugging is kept mostly the same as before. When used as an imported module, logging is less verbose.

See also #101 .

### DIFF
--- a/malsim/__main__.py
+++ b/malsim/__main__.py
@@ -12,6 +12,18 @@ logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
 logging.getLogger().setLevel(logging.INFO)
 
+
+def debug_to_info(record):
+    record.levelname = 'INFO'
+    record.levelno = logging.DEBUG
+    return True
+
+
+module_logger = logging.getLogger('malsim.mal_simulator')
+module_logger.setLevel(logging.DEBUG)
+module_logger.filters.append(debug_to_info)
+
+
 def run_simulation(sim: MalSimulator, agents: list[dict]):
     """Run a simulation with agents"""
 
--- a/malsim/mal_simulator.py
+++ b/malsim/mal_simulator.py
@@ -330,7 +330,7 @@ class MalSimulator():
                 "comes from the agents action surface."
             )
 
-            logger.info(
+            logger.debug(
                 'Attacker agent "%s" stepping through "%s"(%d).',
                 agent.name, node.full_name, node.id
             )
@@ -342,7 +342,7 @@ class MalSimulator():
                 agent.reward += node.extras.get('reward', 0)
                 compromised_nodes.add(node)
 
-                logger.info(
+                logger.debug(
                     'Attacker agent "%s" compromised "%s"(%d).',
                     agent.name, node.full_name, node.id
                 )
@@ -394,7 +394,7 @@ class MalSimulator():
                 "of this simulators attack_graph. Make sure the node "
                 "comes from the agents action surface."
             )
-            logger.info(
+            logger.debug(
                 'Defender agent "%s" stepping through "%s"(%d).',
                 agent.name, node.full_name, node.id
             )
@@ -415,7 +415,7 @@ class MalSimulator():
                     apriori.propagate_viability_from_unviable_node(node)
                 agent.reward -= node.extras.get("reward", 0)
                 enabled_defenses.add(node)
-                logger.info(
+                logger.debug(
                     'Defender agent "%s" enabled "%s"(%d).',
                     agent.name, node.full_name, node.id
                 )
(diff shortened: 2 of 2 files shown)


mrkickling:
I am in favor of this, approving it.

mrkickling (approved):

nkakouros:
What about #101 ?

mrkickling:
We want @andrewbwm s opinion here though, I have not worked much on the logging.

andrewbwm:
I'm quite opposed to this, we're basically just flattening the log level hierarchy for no good reason. I don't understand why you wouldn't just disable logging or set it to WARNING if you wanted to ignore them.

I understand that there are circumstance where we want less logging, what I don't understand is why this would be the preferred solution.

nkakouros:
> I don't understand why you wouldn't just disable logging or set it to WARNING if you wanted to ignore them.

Because I still want some info messages from that module.

> I understand that there are circumstance where we want less logging, what I don't understand is why this would be the preferred solution.

Do you have a simpler solution to suggest? I don't, other than letting every user figure out a solution on their own that will be certainly more complex that this here.

andrewbwm:
> > I don't understand why you wouldn't just disable logging or set it to WARNING if you wanted to ignore them.
> 
> Because I still want some info messages from that module.
> 
> > I understand that there are circumstance where we want less logging, what I don't understand is why this would be the preferred solution.
> 
> Do you have a simpler solution to suggest? I don't, other than letting every user figure out a solution on their own that will be certainly more complex that this here.

As I've already mentioned, I'm perfectly happy with everyone having to figure out a solution on their own. I can't meaningfully predict who wants what type of logging messages.

Just to be clear, I understand why you want it to behave like this, it makes sense. I don't understand why the simulator should match the way you specifically want it to behave as opposed to what other people may want to get as logs.

nkakouros:
> I'm perfectly happy with everyone having to figure out a solution on their own. 

You don't need to make it harder than it has to be though.

> I don't understand why the simulator should match the way you specifically want it to behave as opposed to what other people may want to get as logs.

The PR here changes nothing in the logs that the simulator emits. It only changes log level of some logs for when the simulator is used as a library, logs that are clearly debug level messages, while leaving informational logs intact, such as "sim is starting", "sim is done".

andrewbwm:
> > I'm perfectly happy with everyone having to figure out a solution on their own.
> 
> You don't need to make it harder than it has to be though.
> 
> > I don't understand why the simulator should match the way you specifically want it to behave as opposed to what other people may want to get as logs.
> 
> The PR here changes nothing in the logs that the simulator emits. It only changes log level of some logs for when the simulator is used as a library, logs that are clearly debug level messages, while leaving informational logs intact, such as "sim is starting", "sim is done".



> > I'm perfectly happy with everyone having to figure out a solution on their own.
> 
> You don't need to make it harder than it has to be though.
> 
> > I don't understand why the simulator should match the way you specifically want it to behave as opposed to what other people may want to get as logs.
> 
> The PR here changes nothing in the logs that the malsim command emits. It only changes log level of some logs for when the simulator is used as a library, logs that are clearly debug level messages, while leaving informational logs intact, such as "sim is starting", "sim is done". So, I am not opposing anything, I am just trying to make it easier to cover more cases.



> > I'm perfectly happy with everyone having to figure out a solution on their own.
> 
> You don't need to make it harder than it has to be though.
> 
> > I don't understand why the simulator should match the way you specifically want it to behave as opposed to what other people may want to get as logs.
> 
> The PR here changes nothing in the logs that the malsim command emits. It only changes log level of some logs for when the simulator is used as a library, logs that are clearly debug level messages, while leaving informational logs intact, such as "sim is starting", "sim is done". So, I am not opposing anything, I am just trying to make it easier to cover more cases.

Unless I misunderstood what it does it merge the debug and info levels together. There aren't many other debug messages being sent right now, so it might not be seen as a big issue. But I do not like the design principle behind this.

On a side note, I am getting very tired of these discussions debating what is and what isn't the correct way to log or do some other things. I do not intend to go through them in the future, if nothing else because we need to focus our efforts elsewhere. We will need to come to an agreement about how we continue this work. We can have you decide on these design decisions on the own and we just merge them in or you will likely have to fork the project to implement the changes on your side. We can ask the powers that be which option they prefer, I am happy to go with either one.

nkakouros:
Sounds good, looking forward to this discussion.

mrkickling:
Expectations on logging are probably different for different use cases. It might be worth to think about if you want the default logging at all. I assume you don't run simulations with the cli tool, which means you have full control to do your own logging in your simulation loop that fits your needs (while disabling the default logger, or setting it to 'warning').

nkakouros:
I ended up silencing everything from the mal simulator and adding my own logs.

PR #113, opened 2025-03-05
Initial TTC solution

mrkickling:
- Add method to simulator to sample ttcs (per attacker agent)
  - To think about: should we have this function in mal-toolbox instead?
  - To think about: I took this method from mal-traverser, is it correct?
- Use seed given to Malsimulator.reset so the ttcs are deterministically sampled
- Make sure searcher attacker agents can pick nodes that are not yet compromised
- Only use ttcs in the simulator if MalSimulatorSettings.use_ttcs = True (default False)
  - Supported through cli with `--ttcs` and `--seed <int>`

### DIFF
--- a/malsim/__main__.py
+++ b/malsim/__main__.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 import argparse
 import logging
 
-from .mal_simulator import MalSimulator, MalSimulatorSettings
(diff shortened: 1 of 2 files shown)


malsim/agents/searchers.py
nkakouros, line 1: Can you extract these changes to a separate PR?

PR #114, opened 2025-03-06
Searcher agents allow same node to be selected more than once + forma…

mrkickling:
…tting

- Allow agents to pick a node that was previously picked (if it was not compromised and not removed)
- Make instance variables private (except for .name)
- Make sure .name works for both BFS and DFS
- Comments/Docstrings/variable naming

### DIFF
--- a/malsim/agents/searchers.py
+++ b/malsim/agents/searchers.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 import logging
 import random
-import re
 
 from collections import deque
 from typing import Optional, TYPE_CHECKING, Any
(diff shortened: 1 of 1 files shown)


malsim/agents/searchers.py
```diff
@@ -18,92 +17,100 @@
...
     ) -> Optional[AttackGraphNode]:
+        """Receive the next action according to agent policy (bfs/dfs)"""
+
         self._update_targets(
-            agent_state.step_action_surface_additions
-            if self.started
-            else agent_state.action_surface,
-            agent_state.step_action_surface_removals,
-            **kwargs,
+            new_nodes=(
```
mrkickling, line 55: Just for readability

malsim/agents/searchers.py
```diff
@@ -18,92 +17,100 @@
...
 
-        if self.settings['randomize']:
+        if self._settings['randomize']:
             self.rng.shuffle(new_targets)
 
-        if self.current_target in new_targets:
-            # If self.current_target is not yet compromised, e.g. due to TTCs,
-            # keep using that as the target.
-            new_targets.remove(self.current_target)
-            new_targets.append(self.current_target)
```
mrkickling, line 90: If statement below is longer than I wanted, but I believe it is the most safe way to check if a previous selected target is still available. We could also check if it is in the agent states action surface, but that is not available in this function.

malsim/agents/searchers.py
```diff
@@ -18,92 +17,100 @@
...
-        self.current_target: Optional[AttackGraphNode] = None
-
-        self.settings = self.default_settings | agent_config
+        self._targets: deque[AttackGraphNode] = deque()
+        self._current_target: Optional[AttackGraphNode] = None
+        self._settings = self._default_settings | agent_config
 
-        self.rng = random.Random(self.settings.get('seed'))
-        self.started = False
+        self.rng = random.Random(self._settings.get('seed'))
```
nkakouros, line 46: Why not make rng private too?

malsim/agents/searchers.py
```diff
@@ -18,92 +17,100 @@
...
     def _update_targets(
         self,
-        new_targets_set: set[AttackGraphNode],
-        disabled_nodes_set: set[AttackGraphNode],
-        **kwargs: Any,
+        new_nodes: set[AttackGraphNode],
+        disabled_nodes: set[AttackGraphNode],
     ):
-        if self.settings['seed']:
+        new_targets: list[AttackGraphNode] = []
```
nkakouros, line 73: This does not anything, does it? Initializing the var is not needed, the if/else below will run a branch every time, and mypy should be able to pickup the type.

malsim/agents/searchers.py
```diff
@@ -18,92 +17,100 @@
...
+            # If self.current_target was not compromised, e.g. due to TTCs,
+            # it remains in action surface and should be added as a target.
+            self._targets.append(self._current_target)
 
         # Enabled defenses may remove previously possible attack steps.
-        if disabled_nodes_set:
-            self.targets = deque(n for n in self.targets if n.is_viable)
+        for node in disabled_nodes:
+            if node in self._targets:
+                self._targets.remove(node)
```
nkakouros, line 99: Is this needed to fix a bug? If not this will be slower than the original version.

nkakouros:
Also, I don't understand the title of the PR. Did you add some new functionality that was not there? Does the title mean that the agent should now be able to rerun already compromised steps?

malsim/agents/searchers.py
mrkickling: I missed it

malsim/agents/searchers.py
mrkickling: Yes, I just prefer when variables are initialized in the scope where they are used, but I can remove it.

mrkickling:
> Also, I don't understand the title of the PR. Did you add some new functionality that was not there? Does the title mean that the agent should now be able to rerun already compromised steps?

No, not if it is compromised, but if it has been chosen and not compromised it can be chosen again.
This is solely for the TTC implementation.

malsim/agents/searchers.py
mrkickling: I wanted to remove the nodes from `disabled_nodes_set`, which implicitly should be the same as the nodes that are `not n.is_viable` (so it was not a bug fix).

I didn't realize I made it slower though, it's because deque.remove(node) is O(n) I assume? I will change it back.

nkakouros: Right.

nkakouros (approved):

PR #115, opened 2025-03-06
Add field 'performed_nodes' to agent state containing all perfomed nodes

mrkickling:
For completeness in agent states I think agents should also have this field, so they contain all of the simulations current state.

Before this PR, if you wanted to know what nodes are reached by an attacker you would:

Either:
1. Look at the attacker.reached_attack_steps
  - problematic, since the attacker is replaced after reset(), so you would need to fetch the new attacker from the attack graph
2. Keep track on nodes performed every step and store them somewhere

And to know what nodes are reached by a defender you would:

1. Keep track on nodes performed every step and store them somewhere

Through this PR we give a field in the agent_state containing this information without any extra work having to be done by the user.

### DIFF
--- a/malsim/mal_simulator.py
+++ b/malsim/mal_simulator.py
@@ -34,11 +34,14 @@ class MalSimAgentState:
     # Attackers get positive rewards, defenders negative
     reward: int = 0
 
-    # Contains the steps performed successfully in the last step
-    step_performed_nodes: set[AttackGraphNode] = field(default_factory=set)
-
     # Contains possible actions for the agent in the next step
     action_surface: set[AttackGraphNode] = field(default_factory=set)
+
+    # Contains all nodes that this agent has performed successfully
+    performed_nodes: set[AttackGraphNode] = field(default_factory=set)
+
+    # Contains the steps performed successfully in the last step
+    step_performed_nodes: set[AttackGraphNode] = field(default_factory=set)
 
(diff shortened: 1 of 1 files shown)


nkakouros:
> 1. Look at the attacker.reached_attack_steps
> 
> 
>     * problematic, since the attacker is replaced after reset(), so you would need to fetch the new attacker from the attack graph

Can you explain how this is problematic after a reset()? Isnt it always that you need to access the attacker and its list from the attack graph?

mrkickling:
I guess I would want state information to be returned by step/reset:

```
states = sim.reset()
reached_steps = states['name'].performed_nodes
```

instead of:

```
states = sim.reset()
reached_steps = sim.attack_graph.attackers[attacker.id]
```

I don't want users to have to care about the internal attacker object after the simulator have been initialized.

nkakouros:
> I guess I would want state information to be returned by step/reset:
> 
> ```
> states = sim.reset()
> reached_steps = states['name'].performed_nodes
> ```
> 
> instead of:
> 
> ```
> states = sim.reset()
> reached_steps = sim.attack_graph.attackers[attacker.id].reached_attack_steps
> ```
> 
> I don't want users to have to care about the internal attacker object after the simulator have been initialized.

Ah, I see. Regardless of what you decide to do here, one thing that could be improved is to reference in the agent state the attacker object itself and not just the id.

PR #116, opened 2025-03-07
Make AgentState a read only dataclass rebuilt every time

mrkickling:
This PR introduces a more functional coding style into the simulator. It does not change any functionality and the goal was to not break any of the public APIs, which I hope I didn't.

AgentState objects are not long living states anymore, but only valid for the last step and recreated every time the simulator takes a new step. I think this approach is more elegant in some ways (all agent state changes happen in one place). Basically it gets easier to keep track on the current state of the simulator. 

- Rename MalSimAgentState -> AgentState
- Remove MalSimAgentStateView
  - There is no longer a mutable agent state, just an immutable one

PROS
  - This is safer: nobody will change agent states by mistake
  - This is easier to follow: all changes to the agent state must be done in one place
  - This removes the need for an AgentStateView
  - This is good for the future if we want to let users give states to the simulator to move back/forth to old steps.

CONS:
  - Extra code to rebuild the whole dataclass object every step

### DIFF
--- a/malsim/__main__.py
+++ b/malsim/__main__.py
@@ -15,7 +15,7 @@ logging.getLogger().setLevel(logging.INFO)
 def run_simulation(sim: MalSimulator, agents: list[dict]):
     """Run a simulation with agents"""
 
-    sim.reset()
+    agent_states = sim.reset()
     total_rewards = {agent_dict['name']: 0 for agent_dict in agents}
     all_agents_term_or_trunc = False
 
@@ -38,7 +38,7 @@ def run_simulation(sim: MalSimulator, agents: list[dict]):
                 )
                 continue
 
-            sim_agent_state = sim.agent_states[agent_name]
+            sim_agent_state = agent_states[agent_name]
             agent_action = decision_agent.get_next_action(sim_agent_state)
             if agent_action:
                 actions[agent_name] = [agent_action]
@@ -48,11 +48,11 @@ def run_simulation(sim: MalSimulator, agents: list[dict]):
                 )
 
         # Perform next step of simulation
-        sim.step(actions)
+        agent_states = sim.step(actions)
 
         for agent_dict in agents:
             agent_name = agent_dict['name']
-            agent_state = sim.agent_states[agent_name]
+            agent_state = agent_states[agent_name]
             total_rewards[agent_name] += agent_state.reward
             if not agent_state.terminated and not agent_state.truncated:
                 all_agents_term_or_trunc = False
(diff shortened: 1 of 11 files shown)


nkakouros:
This PR does two things. One is to remove the state view and create new state objects every time. The other is freezing the state and then using new methods and refactoring to cater for this, as freezing was requested by Jakob some time ago.

A comment, using `frozen=True` with dataclass will make the dataclass immutable but not the actual data contained in the dataclass. Mutating state is still possible.

Personally, I don't like the approach. It is a lot of code to do sth that should be simple, i.e. just create a new state object. I also don't think there is a need to change anything. The advantages you mention are already supported, except removing the view class of course which I think you count as an advantage because you didn't like its implementation:



> PROS
> 
>     * This is safer: nobody will change agent states by mistake

It is less safe actually as I described above. Also, safety in Python is a relative concept, people can still mutate state if they want no matter how much you lock this down. I think the important part here is to signal intent and the old view class did that perfectly.

>     * This is easier to follow: all changes to the agent state must be done in one place

Changes in the state where still happening in the a single place, the `step` method. To me, the suggested implementation is harder to follow, I have to follow all these variables and keep them in mind to see, aha, ok, so this variable that we collect here goes to that state property, but it also gets processed a bit in the set_Attack_state method before actually becoming the state prop. And I have to do this for more than one variables.

>     * This is good for the future if we want to let users give states to the simulator to move back/forth to old steps.

The same thing can happen with views.

To sum up, I think this PR is unnecessary.

mrkickling:
> A comment, using frozen=True with dataclass will make the dataclass immutable but not the actual data contained in the dataclass. Mutating state is still possible.

Yes, they could/should be frozenset/etc just like in the AgentStateView if that is what you mean.

> The advantages you mention are already supported, except removing the view class of course which I think you count as an advantage because you didn't like its implementation

I would say that this PR is more about creating a functional solution for the agent states, which makes it clearer (to me at least) where states are created. 

> Changes in the state where still happening in the a single place, the step method. 

I don't agree, they were being made in step, attack_step, defense_step, disable_attack_steps.

> The same thing can happen with views.

Yes, but it is simpler to generate agent states from an attack graph with this solution I would say.

> To sum up, I think this PR is unnecessary.

Alright. I don't think it is necessary, but I was hoping I could raise the interest for a more functional approach.

andrewbwm:
My general opinion regarding such things is that we should leave them for after we're done with the current project.

mrkickling:
Not for now

PR #121, opened 2025-03-20
Create methods to calculate rewards and termination and use those instead of calculating in place

mrkickling:
Create `_attacker_step_reward`, `_defender_step_reward`, `_attacker_is_terminated`, `_defender_is_terminated`.

This came up after discussions in the MAL Plenary, and the reasoning is to make it easier for users to override this logic in particular instead of rewriting bigger part of the logic. (@sandorstormen is the one who brought it up).

As a bonus it also makes it clearer how rewards are set each step.
Before they were set in place in attacker_step and defender_step, but now the method is called in step.

I think we should do the same for all parts of the agent state calculations but I kept this PR limited to rewards and termination.

### DIFF
--- a/malsim/mal_simulator.py
+++ b/malsim/mal_simulator.py
@@ -226,21 +226,21 @@ class MalSimulator():
         """Give rewards for pre-enabled attack/defense steps"""
 
         pre_enabled_attack_steps = set()
-        pre_enabled_defenses = set(
-            node for node in self.attack_graph.nodes.values()
-            if node.is_enabled_defense()
-        )
-
         for attacker_state in self._get_attacker_agents():
+            # Attackers get positive reward for pre-compromised nodes
             attacker = attacker_state.attacker
-            attacker_state.reward = self._attacker_step_reward(
+            attacker_state.reward = self._attacker_reward(
                 attacker_state, attacker.reached_attack_steps,
             )
             pre_enabled_attack_steps |= attacker.reached_attack_steps
 
+        pre_enabled_defenses = set(
+            node for node in self.attack_graph.nodes.values()
+            if node.is_enabled_defense()
+        )
         for defender_state in self._get_defender_agents():
-            # Defenders get negative reward for pre-enabled nodes
-            defender_state.reward = self._defender_step_reward(
+            # Defenders get negative reward for pre-enabled/compromised nodes
+            defender_state.reward = self._defender_reward(
                 defender_state, pre_enabled_defenses, pre_enabled_attack_steps
             )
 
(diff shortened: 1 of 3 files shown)


tests/envs/test_vectorized_obs_mal_simulator.py
```diff
@@ -200,62 +200,6 @@ def test_create_blank_observation_actionability_given(
         else:
             assert not actionable
 
-def test_step(corelang_lang_graph, model):
```
mrkickling, line 203: This was already tested in test_mal_simulator.py

sandorstormen (commented):
Maybe there should be some documentation that `_attacker_step_reward`, `_defender_step_reward`, `_attacker_is_terminated` and `_defender_is_terminated` are functions that you could override? Otherwise it looks good to me.

malsim/mal_simulator.py
```diff
@@ -487,6 +481,39 @@ def _defender_step(
...
+        )
+
+    def _defender_step_reward(
+            self,
+            defender_state: MalSimDefenderState,
+            step_enabled_defenses: set[AttackGraphNode],
+            step__compromised_nodes: set[AttackGraphNode]
+        ):
+        """Defender is penalized for compromised nodes and enabled defense"""
+        return defender_state.reward - sum(
```
kasanari, line 504: Why is the reward a function of the reward? The reward for a state transition should only depend on the state and action.

kasanari: the function does not need self

mrkickling: Clarified in docstring what the method does.
Made method static.

nkakouros:
How do you imagine the user will override these new methods? Will they subclass the simulator and override them? A different approach that I think would be easier to follow for users is for the simulator to use a `MalSimulatorPolicy` object (or other name) that provides methods for reward calculation etc and this is passed as a dependency to the mal simulator.

malsim/mal_simulator.py
andrewbwm: I mean, it doesn't matter that much, but I'm not a fan of static methods like this. I am ok with them taking self or at least cls as a parameter even if they don't make modifications. But leave it as is now.

andrewbwm:
> How do you imagine the user will override these new methods? Will they subclass the simulator and override them? A different approach that I think would be easier to follow for users is for the simulator to use a `MalSimulatorPolicy` object (or other name) that provides methods for reward calculation etc and this is passed as a dependency to the mal simulator.

I'm not necessarily sure that's "easier" and we can really ponder this more when we do the next re-design once we've moved the `Attacker` into the Simulator. For the time being this will suffice.

andrewbwm (approved):

PR #122, opened 2025-03-26
Handle rewards as regular scenario properties

andrewbwm:
- [ ] Handle Rewards as scenario properties
- [ ] Add scenario converter file to update rewards format
- [ ] Merge scenario property rules and values into a single concept


malsim/scenario.py
```diff
@@ -158,56 +159,16 @@ def _validate_scenario_node_property_config(
...
-                step.extras[node_prop] = 0
-
-
-def apply_scenario_node_property_values(
-        attack_graph: AttackGraph, node_prop: str, values: dict,
+def apply_scenario_node_property(
+        attack_graph: AttackGraph,
+        node_prop: str,
+        prop_conf: dict,
+        prop_default: Optional[Any] = None,
```
mrkickling, line 170: This is also a `value_default`, that only applies when `prop_conf` is empty.
Try to find a variable name that expresses this better.

malsim/scenario.py
```diff
@@ -216,43 +177,89 @@ def apply_scenario_node_property_values(
...
+                        nodes if property is entirely omitted in the
+                        configuration. If None no values will be set.
+    - value_default:    the default value to set for the property for nodes
+                        where no value is given in the configuration. If None
+                        no values will be set. This is only relevant if the
+                        property is included in the scenario configuration.
     """
 
-    _validate_scenario_node_property_config(attack_graph, values)
+    def _extract_value_from_entries(entries: [dict|list], step_name: str) -> Any:
```
mrkickling, line 193: typehint for entries should be 'dict|list' instead of '[dict|list]'

malsim/scenario.py
```diff
@@ -216,43 +177,89 @@ def apply_scenario_node_property_values(
...
+            return value
+        else:
+            raise ValueError('Error! Scenario node property configuration '
+                'is neither dictionary, nor list!')
+
+    _validate_scenario_node_property_conf(attack_graph, prop_conf)
 
-    if not values:
+    if not prop_conf:
         # If no rules defined, do not set the values at all
```
mrkickling, line 221: Should maybe be updated to:  'If no conf defined for property, set the attribute value for each node to `value_default` if it is set, then return.'

malsim/scenario.py
```diff
@@ -216,43 +177,89 @@ def apply_scenario_node_property_values(
...
+            for step in attack_graph.nodes.values():
+                step.extras[node_prop] = prop_default
         return
+    else:
+        if value_default is not None:
+            for step in attack_graph.nodes.values():
+                step.extras[node_prop] = value_default
 
     for step in attack_graph.nodes.values():
         # See if user defined rule that sets the specified
```
mrkickling, line 232: as you notice, most of my comments are naming related, but here i think we need to decide if we want to call them rules. If we want to call them rules the variable name should also be 'prop_rules'

andrewbwm: No, we don't want to call them rules because that doesn't mean anything. You are right however, I did miss this comment. :)

PR #124, opened 2025-04-08
Update README, test and raise specific error for wrong rewards format

mrkickling:
Reward format was updated for scenario file.
To make the transition easier:
- Raise exception with information
- Update README


mrkickling:
This should be merged before next release, since it gives info about the breaking change for how users set rewards.

andrewbwm (approved):

PR #128, opened 2025-05-28
Change where agent states are created and updated

mrkickling:
- all agent states are now created in one place _create_attacker_state/_create_defender_state
- all agent states are updated in one place to avoid confusion
- added initial action surface to agent_state.attack_surface_additions on creation

### DIFF
--- a/malsim/mal_simulator.py
+++ b/malsim/mal_simulator.py
@@ -222,60 +222,137 @@ class MalSimulator():
 
         return self.agent_states
 
-    def _init_agent_action_surfaces(self):
-        """Set agent action surfaces according to current state"""
-        for agent in self._agent_states.values():
+    def _create_attacker_state(
+        self, name: str, attacker: Attacker
+    ) -> MalSimAttackerState:
+        """Create a new defender state, initialize values"""
+        attacker_state = MalSimAttackerState(name, attacker)
+        attacker_state.step_performed_nodes = attacker.reached_attack_steps
+        attacker_state.performed_nodes = attacker.reached_attack_steps
+        attacker_state.action_surface = query.calculate_attack_surface(
+            attacker, skip_compromised = True
+        )
+        attacker_state.step_action_surface_additions = (
+            set(attacker_state.action_surface)
+        )
+        attacker_state.step_action_surface_removals = set()
+        attacker_state.reward = self._attacker_reward(attacker_state)
+        return attacker_state
 
-            if isinstance(agent, MalSimAttackerState):
-                attacker = agent.attacker
-                agent.action_surface = query.calculate_attack_surface(
-                    attacker, skip_compromised = True
-                )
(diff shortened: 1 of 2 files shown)


malsim/mal_simulator.py
```diff
@@ -556,57 +600,54 @@ def step(
...
+                self._update_defender_state(
+                    agent_state,
+                    step_all_compromised_nodes,
+                    step_enabled_defenses,
+                    step_nodes_made_unviable
+                )
+            else:
+                raise TypeError(
+                    f"Unknown agent class {agent_state.__class__}"
+                )
```
nkakouros, line 645: I don't think this branch is needed. The only ones that handle the self._agent_states variable is us in this class. It is private and users of this class should not touch it. If they do, their problem.

malsim/mal_simulator.py
```diff
@@ -222,60 +222,137 @@ def reset(
...
-        for agent in self._agent_states.values():
-
-            if isinstance(agent, MalSimAttackerState):
-                attacker = agent.attacker
-                agent.action_surface = query.calculate_attack_surface(
-                    attacker, skip_compromised = True
-                )
+    def _create_attacker_state(
+        self, name: str, attacker: Attacker
+    ) -> MalSimAttackerState:
```
nkakouros, line 227: Splitting into create and update seems redundant. We could keep the latter only. When a new state is needed, we just create the object and pass that in the update method with the rest of the args being empty sets. The same for the defender.

If you do the above you may feel that creating manually a state object in register_attacker/register_defender and reset_agents does not look nice. The solution is for register_* to call reset_agents, until Andrei removes the Attackers completely. Then we can revisit if needed.

nkakouros (commented):
The changes here make sense overall.

malsim/mal_simulator.py
```diff
@@ -556,57 +600,54 @@ def step(
...
+                    step_nodes_made_unviable
+                )
+
+            elif isinstance(agent_state, MalSimDefenderState):
+                self._update_defender_state(
+                    agent_state,
+                    step_all_compromised_nodes,
+                    step_enabled_defenses,
+                    step_nodes_made_unviable
+                )
```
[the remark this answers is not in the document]
mrkickling: The reason for doing it this way was so I could remove 'dead' agents in one loop instead of having to duplicate it. 

Now I will try this approach (every list item is a loop):
1. Step with defenders
2. Step with attackers and update their states
3. Update defender states and remove any 'dead' agents

The previous approach was to also update attacker states in the last loop, but still we need to have a loop to remove dead agents (could be a separate one, but I think this will be fine)

malsim/mal_simulator.py
andrewbwm: I agree with what you both said and I'll re-evaluate it after the attacker logic has been moved and I see what it looks like. The reason I will leave it in for now is this blank state is needed in two places, when the simulator is created and upon reset.

PR #134, opened 2025-06-30
FP for Gym interface

mrkickling:
- [x] Add code examples on how to use the different interfaces programmatically.
- [x] Implement false positives
- [x] Test the _generate_false_positives method
- [x] Test false positives actually being present in observations in the serialized obs env / gym envs

Branch made for Yacine - work in progress.


README.md
```diff
@@ -219,3 +219,168 @@ options:
...
+
+# info contains serialized action mask
+obs, info = env.reset()
+
+# Simulation loop
+term = False
+while not term:
+    # Action selection should not be handled like this is you
+    # are training an ML agent naturally
+    defender_name = env.unwrapped.defender_agent_name
```
kasanari, line 376: You could use env.observation_space.sample(action_mask) to not have to use mal-simulator-specific attributes in the example.

PR #136, opened 2025-08-01
Add feature "Extend scenarios"

mrkickling:


tests/testdata/scenarios/traininglang_scenario_extended.yml
```diff
@@ -0,0 +1,15 @@
...
+    Host:0:
+      notPresent: 1
+      access: 1
+    Host:1:
+      notPresent: 1
+      access: 1
+    Data:2:
+      notPresent: 1
+      read: 1
+      modify: 1
```
andrewbwm, line 15: None of the tests actually test that an override has occurred, they just add new values, they never override existing ones.

PR #147, opened 2025-08-21
Implement TTCs

andrewbwm:
- [x] Add probabilities utility module to handle all probability calculations
- [x] Implement a Bernoulli dictionary to ensure that Bernoulli sampling only happens once
- [x] Update necessity and viability logic to work with Bernoulli results
- [x] Add probability mode option to `MalSimulatorSettings` that specifies how to use the TTC distribution of a node
- [x] Implement probability mode in attack step resolution
- [x] Add seed to `MalSimulatorSettings` to use to make random number generator deterministic
- [x] Update environments to work with the TTC processing
- [x] Update tests to work with the TTC processing
- [x] Update code to be mypy compliant

Unrelated:
- [x] a basic test model that is cleaner than the old simple test model, which I think we should switch to entirely
- [x] update `mal-toolbox` requirement to `1.*`

### DIFF
--- a/malsim/mal_simulator.py
+++ b/malsim/mal_simulator.py
@@ -22,7 +22,6 @@ from malsim.probs_utils import (
 from malsim.graph_processing import (
     calculate_viability_and_necessity,
     propagate_viability_from_node,
-    prune_unviable_and_unnecessary_nodes
 )
 
 ITERATIONS_LIMIT = int(1e9)
@@ -174,7 +173,6 @@ class MalSimulatorSettings():
     #   traversable (often because a defense kicked in) if set to True
     # otherwise:
     # - Leave the node/step compromised even after it becomes untraversable
-    prune_unviable_unnecessary: bool = True
     uncompromise_untraversable_steps: bool = False
     prob_mode: ProbMode = ProbMode.SAMPLE
     seed: Optional[int] = None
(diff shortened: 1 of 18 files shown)


malsim/mal_simulator.py
```diff
@@ -159,7 +174,10 @@ class MalSimulatorSettings():
     #   traversable (often because a defense kicked in) if set to True
     # otherwise:
     # - Leave the node/step compromised even after it becomes untraversable
+    prune_unviable_unnecessary: bool = True
     uncompromise_untraversable_steps: bool = False
+    prob_mode: ProbMode = ProbMode.SAMPLE
```
mrkickling, line 179: TTC_mode is clearer if it only relates to TTCs

malsim/mal_simulator.py
```diff
@@ -202,54 +214,52 @@ def __init__(
         self._viable_nodes: set[AttackGraphNode] = set()
```
mrkickling, line 214: maybe we should call a function for each of these attributes, instead of setting them to empty set and perform "prepare_attack_graph" that updates them all. It would be clearer. Not important but it would be neater I think.

malsim/probs_utils.py
```diff
@@ -0,0 +1,197 @@
...
+
+from typing import Any, Optional
+
+logger = logging.getLogger(__name__)
+
+class ProbCalculationMethod(Enum):
+    SAMPLE = 1
+    EXPECTED = 2
+
+bernoulli_dict: dict[int, float] = {}
```
mrkickling, line 16: This one should be in the simulator to make it easier to find, and we don't need the clear_bernoulli_dict function but can just run dict.clear in the simulator when we clear it

mrkickling (approved):

PR #149, opened 2025-08-27
Remove extras improve scenarios

mrkickling:
Making scenario a dataclass, storing rewards in a separate dictionary instead of using the magic dict node.extras. Added class method to create sim from scenario.

Might also make sense to have automatic agent registering when loading sim from scenario, any ideas on that?


mrkickling:
Anyone has any thoughts here? Making scenario a dataclass, storing rewards in a separate dictionary instead of node.extras. I think they are improvements.

kasanari:
I approve :+1:

kasanari (approved):

mrkickling:
Thank you everyone for the approvals.

PR #167, opened 2025-09-15
Immutable agent state

mrkickling:
In current main, `MalSimAgentState` stores all info about agents.
These objects are mutable in the simulator, but the simulator returns `MalSimAgentStateView` to users which are read only projections of the agent states.

In this PR I have made MalSimAgentState immutable. This means that they are recreated in every step based on what happened in that step. This also removes the need of `MalSimAgentStateView`

Reward and termination is not stored in the agent state, but instead in the simulator. 

Some additions is also that I added `.done()` which can tell a user when the simulation is done (can be used in their simulation loop). I also added some other helpers and split up some code to make the immutable agent states work and also just because it looked nice.

Some of the changes are just moving methods around in the mal_simulator module into a more logical order.

Please take a look and see what you think about this PR.

### DIFF
--- a/malsim/agents/decision_agent.py
+++ b/malsim/agents/decision_agent.py
@@ -5,7 +5,7 @@ from typing import TYPE_CHECKING, Optional, Any
 from abc import ABC, abstractmethod
 
 if TYPE_CHECKING:
-    from ..mal_simulator import MalSimAgentStateView
+    from ..mal_simulator import MalSimAgentState
     from maltoolbox.attackgraph import AttackGraphNode
 
 class DecisionAgent(ABC):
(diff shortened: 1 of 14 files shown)


sandorstormen:
It's still a bit weird to me that `truncation` is in the `mal-simulator` since most RL-frameworks have their own truncation logic. But I guess it's fine to set it to a big number as a work around.

mrkickling:
> It's still a bit weird to me that `truncation` is in the `mal-simulator` since most RL-frameworks have their own truncation logic. But I guess it's fine to set it to a big number as a work around.

Yeah, it is set to some

mrkickling:
> It's still a bit weird to me that `truncation` is in the `mal-simulator` since most RL-frameworks have their own truncation logic. But I guess it's fine to set it to a big number as a work around.

We could just get rid of the term truncation in the simulator, and instead say that simulation is 'done' on certain conditions (agents have no actions left or max iter is reached). The term itself is not really important.

PR #168, opened 2025-09-15
Use numpy.rng instead of std random for ttc calcs, update tests

mrkickling:


sandorstormen:
What's the thinking behind this? Is numpy faster or something?

PR #178, opened 2025-09-19
Add initial solution with ttc values as attacker rewards

mrkickling:


sandorstormen:
Can I set `ttc_values_as_attacker_penalty` to `True` and still select how TTC values are sampled?

mrkickling:
No, an attacker agent is not penalized with 1 point per step it takes, and never was. It is not penalized at all as it is now.

mrkickling:
How should we move forward with this in your opinion?

sandorstormen:
Give me some time and I can implement it.

PR #189, opened 2025-09-29
Full name support

mrkickling:


kasanari:
Would these changes also apply to the configurations dicts for FPR and so on:
```
observability_per_node: Optional[dict[AttackGraphNode, bool]] = None,
actionability_per_node: Optional[dict[AttackGraphNode, bool]] = None,
false_positive_rates: Optional[dict[AttackGraphNode, float]] = None,
false_negative_rates: Optional[dict[AttackGraphNode, float]] = None,
```
they currently seem to be only using the node object.

mrkickling:
No, but that might be a good idea

mrkickling:
Now it is possible

PR #208, opened 2025-10-14
Ttcs per attacker agent

mrkickling:
Make it possible to give attacker agents specific TTCs.
So each attacker could set TTCs for attack steps per asset type or asset name.

Reasoning:
- Some attackers might be better at certain attacks
- Could mimic certain threat actors with certain strengths

In progress, trying to figure out how we want users to set ttcs in scenario (strings, dicts..)


mrkickling:
This one needs rebasing and to be added into the new scenario functionality. Will do.

PR #219, opened 2025-10-22
Stricter Type Checking for TTCMode

sandorstormen:


malsim/mal_simulator.py
```diff
@@ -174,6 +174,14 @@ def infinity_ttc_attack_step_unviable(self) -> None:
             "'run_attack_step_bernoullis'"
         )
 
+    def __post_init__(self) -> None:
```
mrkickling: You want to be able to give the ttc mode as a string?

PR #220, opened 2025-10-23
Graphical Gym Interface with Language Serialization

sandorstormen:
Implementation of a new Gym-like interface for the simulator. The observation is now a graph-object which includes different MAL constructs.

- Attack Steps
- Assets
- Association
- Logic Gates (Optional)
  - The logic gates are node-representations of `AND` and `OR` operations for attack steps.

Lang graph representation of the `MalObsInstance`:
GraphicalGymInterface drawio

Each construct has some associated attributes.
- Attack Steps
  - Type: Attack step name (the Type of asset can also be included here)
  - ID: Node id
  - Class: The logic class $\in [\texttt{AND}, \texttt{OR}, \texttt{DEFENSE}, \texttt{Exists}, \texttt{notExists}]$
  - Tag: The @-tag of the step
  - Compromised: Whether the step is compromised
  - Traversable: Weather the step is traverable
  - Number of Attempts: Number of times this attack step has been attempted
- Assets
  - Type: Asset name
  - ID: Asset id
- Association
  - Type: Association name (The types of associations can optionally be included here.)
- Logic Gates
  - ID: The ID of the attack step, for which the class the logic gate represents. 
  - Type: `AND` or `OR`

### DIFF
--- a/malsim/envs/graph/mal_spaces.py
+++ b/malsim/envs/graph/mal_spaces.py
@@ -9,12 +9,12 @@ from .serialization import LangSerializer
 from malsim.mal_simulator import MalSimulator
 
 
-class Asset(NamedTuple):
-    """An asset in the attack graph.
+class Assets(NamedTuple):
+    """Assets in the attack graph.
 
     Attributes:
-        type: The type of the asset (AssetType).
-        id: The ID of the asset in the model.
+        type: The type of the assets (AssetType).
+        id: The IDs of the assets in the model.
         action_mask: Whether there is any step connected to the asset which can be
             performed (i.e. step.action_mask == True).
 
@@ -25,19 +25,19 @@ class Asset(NamedTuple):
     action_mask: NDArray[np.bool_]
 
 
-class Step(NamedTuple):
-    """A node in the attack graph.
+class Steps(NamedTuple):
+    """Steps in the attack graph.
 
     Attributes:
-        type: The type of the step (AssetType, AttackStepName) or (AttackStepName,).
-        id: The ID of the step in the attack graph.
+        type: The type of the steps (AssetType, AttackStepName) or (AttackStepName,).
+        id: The IDs of the steps in the attack graph.
         logic_class: One of {'and', 'or', 'defense', 'exist', 'notExist'}.
-        tags: The first @-tag of the step.
-        compromised: Whether the step is observed as compromised.
-        observable: Whether the step can be observed to be compromised.
-        attempts: The number of times the step has been performed.
+        tags: The first @-tags of the steps.
+        compromised: Whether the steps are observed as compromised.
+        observable: Whether the steps can be observed to be compromised.
+        attempts: The number of times the steps have been performed.
             Only relevant for the attacker.
-        action_mask: Whether the step can be performed.
+        action_mask: Whether the steps can be performed.
     """
 
     type: NDArray[np.int64]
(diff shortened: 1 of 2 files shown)


malsim/envs/serialization.py
```diff
@@ -0,0 +1,68 @@
...
+        """
+        self._lang = lang
+        self.split_assoc_types = split_assoc_types
+        self.split_attack_step_types = split_attack_step_types
+
+        # TODO: Remove abstract classes?
+        self.asset_type: dict[str, int] = {
+            asset_type: i for i, asset_type in enumerate(sorted(self._lang.assets.keys()))
+        }
+        self.association_type: dict[str, dict[str, dict[str, int]]] | dict[str, int] = {}
```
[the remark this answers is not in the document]
mrkickling: okay, so if you 'split_assoc_types' then it will be str to dict, otherwise str to int..

malsim/envs/serialization.py
```diff
@@ -0,0 +1,68 @@
...
+        Args:
+            lang: LanguageGraph to serialize
+            split_assoc_types: If True, split association types into separate types for each asset type
+        """
+        self._lang = lang
+        self.split_assoc_types = split_assoc_types
+        self.split_attack_step_types = split_attack_step_types
+
+        # TODO: Remove abstract classes?
+        self.asset_type: dict[str, int] = {
```
mrkickling: asset_type_to_idx instead perhaps?

mrkickling: asset_type_idx then? I will not fight for it more if you disagree but I think it would be easier to understand

malsim/envs/graph/serialization.py
```diff
@@ -0,0 +1,96 @@
...
+        # LanguageGraphAssociationField attributes minimum and maximum
+
+        all_attack_steps = sorted([
+            attack_step for asset in self._lang.assets.values()
+            for attack_step in asset.attack_steps.values()
+        ], key=lambda x: x.name)
+
+        # NOTE: This looks odd, but the attack step name is the type of the action
+        self.attack_step_type: dict[tuple[str, ...], int] = {}
+        if split_attack_step_types:
```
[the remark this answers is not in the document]
sandorstormen: It's not really an unholy variable. It's just an option to have Vejde-style types (AssetType.StepName) or to have semantic types (StepName). Without splitting the attack step types, you can still know about their relation to the asset type by their relation to the assets.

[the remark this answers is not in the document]
kasanari: This is what the [`unholy` variable used to control](https://github.com/mal-lang/mal-simulator/commit/1b7058006ddced4d260ab61468b71f8df07d5733), before it was removed.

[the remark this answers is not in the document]
sandorstormen: Aha, then I have misunderstood the concept of `unholy`.

My thinking is just that a step that is named `fullAccess`, for example, will function similarly if it's applied to `Hardware` or `Application`. If that is not the case with the language of generalization, you set `split_attack_step_types=True`.

malsim/envs/graph/README.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Action Spaces
+
+The interface supports multiple action space representations:
+
+#### Step-Based Action Spaces
+
+- **`MALObsAttackStepSpace`**: Discrete space over actionable attack steps (AND/OR nodes)
+  - Actions are indices into the sorted list of actionable attack steps
+  - Assumes all `and`/`or` steps have lower indicies than `defense`/`exist`/`notExist`
```
mrkickling: is this an assumption on the simulator/toolbox or on the MalObsInstance?

malsim/envs/graph/README.md
```diff
@@ -0,0 +1,174 @@
...
+
+- **`MALObs`**: Base observation space
+- **`MALAttackerObs`**: Attacker observation space (requires `attempts` field)
+- **`MALDefenderObs`**: Defender observation space (must NOT have `attempts` field)
+
+### Environment Types
+
+#### Single-Agent Environments
+
+- **`MalSimAttackerGraph`**: Single-agent attacker environment
```
mrkickling: MalSimAttackerGraph sounds like a graph, but it is actually an env, right? AttackerGraphEnv + DefenderGraphEnv?

malsim/envs/graph/graph_env.py
```diff
@@ -0,0 +1,329 @@
...
+    ) -> tuple[MALObsInstance, dict[str, Any]]:
+        super().reset(seed=seed, options=options)
+        obs, info = self.multi_env.reset(seed=seed, options=options)
+        return obs[self.agent_name], info[self.agent_name]
+
+    def step(
+        self, action: np.int64
+    ) -> tuple[MALObsInstance, SupportsFloat, bool, bool, dict[str, Any]]:
+        obs, reward, terminated, truncated, info = self.multi_env.step(
+            {self.agent_name: action}
```
mrkickling: No opponents?

malsim/envs/graph/README.md
mrkickling: Yes, that is better, I would be fine with AttackerGraphEnv as well, but maybe it feels to general for you.

PR #225, opened 2025-10-26
Add penalties for attacker

sandorstormen:
Uses TTC value as a cost for the attacker. When TTC is disabled, TTC value is either sampled or the expectation is taken. When TTC is enabled, the cost is proportional to the amount of failed steps.

### DIFF
--- a/malsim/mal_simulator.py
+++ b/malsim/mal_simulator.py
@@ -1104,11 +1104,11 @@ class MalSimulator():
             # If TTC Mode is disabled but reward mode uses TTCs, penalize attacker with TTCs
             for node in attacker_state.step_performed_nodes:
                 if self.sim_settings.attacker_reward_mode == RewardMode.EXPECTED_TTC:
-                    step_reward -= ttc_value_from_node(node, ProbCalculationMethod.EXPECTED, self.rng)
+                    step_reward -= ttc_value_from_node(node, ProbCalculationMethod.EXPECTED, self.rng) if node.ttc else 0
                 elif self.sim_settings.attacker_reward_mode == RewardMode.SAMPLE_TTC:
-                    step_reward -= ttc_value_from_node(node, ProbCalculationMethod.SAMPLE, self.rng)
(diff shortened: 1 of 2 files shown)


malsim/mal_simulator.py
```diff
@@ -1095,6 +1097,18 @@ def _attacker_step_reward(
             for n in attacker_state.step_performed_nodes
         )
 
+        if self.sim_settings.ttc_mode in (TTCMode.EFFORT_BASED_PER_STEP_SAMPLE, TTCMode.PER_STEP_SAMPLE, TTCMode.EXPECTED_VALUE, TTCMode.PRE_SAMPLE):
```
mrkickling: We don't check this in ruff, but I think this line is too long.
Suggestions:

```python

        non_disabled_ttc_modes = (
            TTCMode.EFFORT_BASED_PER_STEP_SAMPLE,
            TTCMode.PER_STEP_SAMPLE,
            TTCMode.EXPECTED_VALUE,
            TTCMode.PRE_SAMPLE
        )
        if self.sim_settings.ttc_mode in non_disabled_ttc_modes:
            step_reward -= len(attacker_state.step_attempted_nodes)
        elif self.sim_settings.ttc_mode == TTCMode.DISABLED:
            for node in attacker_state.step_attempted_nodes:
                if self.sim_settings.attacker_reward_mode == RewardMode.EXPECTED_TTC:
                    step_reward -= ttc_value_from_node(
                        node, ProbCalculationMethod.EXPECTED, self.rng
                    )
                elif self.sim_settings.attacker_reward_mode == RewardMode.SAMPLE_TTC:
                    step_reward -= ttc_value_from_node(
                        node, ProbCalculationMethod.SAMPLE, self.rng
                    )
                else:
                    raise ValueError(
                        f"Invalid RewardMode when TTC mode is DISABLED: {reward_mode}"
                    )


```

Or you could just do `if self.sim_settings.ttc_mode != TTCMode.DISABLED` since that seems to be what we are interested in here.

PR #226, opened 2025-10-27
Normalize ttc solutions

mrkickling:
Merge the TTC solutions to work with the same objects and use scipy built in distribution functions.
Simplifies things and unifies the TTC implementations.

Still give options to use:
- Expected value
- Sampled value
- Sampled value + effort

Needs a bit more testing with a language that has more advanced TTCs set.


sandorstormen (approved):
LGTM

PR #231, opened 2025-10-29
ref: shorter lines and tuple keys instead of dicts in serialization

mrkickling:
- Store split assocations/attack steps with tuples as keys maping to idx instead of layered dicts
- Shorter lines in code base, split up lines where possible


sandorstormen (approved):
LGTM

Maybe change the comment to be explicit about the left/right asset type having index 2 and 3, resp. in the tuple.

PR #235, opened 2025-10-30
feat: Use actionability to limit agent action surfaces

mrkickling:
This was previously just implemented in Jakobs interface, but it makes sense that the result of being actionable/non actionable is to be or not to be in the action surface of any agent.

Note:
The annoying thing with actionability is that you would have to make every single type of node actionable except for one if you want to make one node non-actionable.

Actionability is also not set by agent type, so you would need to add actionability for both defense and attack steps even though you are just working from defender or attacker perspective and want the other type of agent to have full actionability.

This makes actionability kind of shitty. Any ideas?

### DIFF
--- a/malsim/mal_simulator.py
+++ b/malsim/mal_simulator.py
@@ -628,9 +628,7 @@ class MalSimulator:
                     and not self.node_is_necessary(child)
                 ):
                     continue
-                if (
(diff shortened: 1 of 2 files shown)


sandorstormen (approved):
LGTM

PR #236, opened 2025-10-30
feat: Add option to not compromise entry points initially

mrkickling:


kasanari (approved):
The option works as a temp solution, but as mentioned in the original issue I think allowing the initial state to be set is the cleaner solution in the long run. 

I think there is already some danger in the state-creation logic being separated between _create and _update, so adding additional differences is making me a bit uneasy.

PR #239, opened 2025-11-10
feat: Add Order of Performed Nodes

sandorstormen:
Changes `performed_nodes` to immutable ordered sequence.


kasanari:
Would adding temporal info (e.g. timestep or similar) to the compromised attack steps not be a more general solution, in case the order or temporal location of the events are of interest?

mrkickling (commented):
Looks good. Only problem I have is that the method `node_is_compromised` will go from finding a node in a set to finding a node in a tuple which means much worse performance. That function is not used internally so it doesn't affect performance in the simulator, but it could be used every step by some of the wrappers/interfaces.

PR #258, opened 2025-12-02
Never sample degenerate distributions.

sandorstormen:
Degenerate distributions shouldn't be sampled. If a defense property is set on an asset, that should be respected, instead of being turned of.


sandorstormen:
https://en.wikipedia.org/wiki/Degenerate_distribution

PR #265, opened 2025-12-18
Make the cumulative option work as expected

kasanari:
In a well-defined MDP the reward should only be based on the current state and action, so it would be better to not default to the non-markovian option.

# Update
I realized that the function was calculating things in a way I do not fully grasp, so rather than changing the option I changed it to work like expected.

### DIFF
--- a/malsim/mal_simulator/simulator.py
+++ b/malsim/mal_simulator/simulator.py
@@ -160,7 +160,7 @@ class MalSimulator:
             sim_settings.defender_reward_mode
         ]
 
-    def __getstate__(self):
+    def __getstate__(self) -> dict[str, Any]:
         do_not_pickle = {
             'performed_attacks_func',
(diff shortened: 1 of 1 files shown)


kasanari:
General comment regarding the tests: there seems to be a tendency for tests to have assertions not related to the property being tested. For instance, `test_attacker_step_rewards_cumulative` has an assertion about the recording being in a certain order. Is there a reason for doing this rather than having separate tests for these different properites?

kasanari:
At this point the only test failing is `test_simulator_picklable`, which fails because pickle can not serialize lambdas. However, I would argue this is not a feature that should be supported anyway, at least in its current implementation. I am unsure what the purpose of it is, but if it is to be able to save the state of the simulator, then it would be better that the state is stored in a separate class which is serialized.

PR #266, opened 2025-12-18
Move reward function out of simulator

kasanari:
To avoid accidental leakages of statefulness.

Should also make it easier to define custom reward functions.


kasanari:
This PR is also outdated since I realized the reward functions were not behaving as I expected.

PR #267, opened 2025-12-19
Remove max iter and put iterations in the state

kasanari:
This PR does two things.

- Remove max iters. I think the use of this parameter has been misinterpreted, so it is safer to not have the simulator handle it.
- Move the current iteration into the state. The current iteration is stateful, so it should be maintained in the state. If it makes more sense that time is shared in the simulator, there should be an underlying sim state class.


kasanari:
Also, I can agree that having some common state might make sense, but then that state should be probably be stored in some kind of struct, e.g. SimState

Issue #86, opened 2024-12-16
Automatically traverse attack steps that are zero cost and reachable


mrkickling:
Requested by @Hoclor 

Problem:
- AttemptX/SuccesfulX/X is not very intuitive
- Instead the attacker should automatically perform the 'hidden' attack steps automatically before finishing a step in the simulator.
- The simulator then returns the state.

This might be related to #84 since it should only happen if the ttc is 0 (zero cost).

Please add any discussion details.

kasanari:
> For example if steps A, B and C are all 0 TTC, and D is TTC 1, and the attacker attempts A, should we do A-B-C, or A-B-C-D?

If A, B, C have TTC=0, and we short-circuit, the attacker would never be able to attempt A since it is already compromised as soon as the preconditions of A are fulfilled, so D would be the choice in the chain A-B-C-D. If A has TTC=1,B,C=0 and D=1, then the only possible action sequence is [A,D].

kasanari:
I am deliberately avoiding speculating about features that are currently not in the simulator logic.

> What is the short-circuited steps have detectors on them when should these trigger? We can argue that the user should not define detectors on zero TTC steps, but we would be juggling semantics and simulator implementation to make things works. Also, TTC==0 may be a runtime result, not one defined in the language.

I do not know how detectors are intended to work, so I can not say how steps auto-firing would impact them. I would think the step being observed is _because_ there is a detector on it. TTC=0 in my mind is any exogenous state variable change that might happen, so if these are observed or not is up to the model I suppose.

With 'the language' I assume you mean MAL. The simulator is already using TTC's in a way that is not in line with how MAL defines them. MAL defines TTC's as either a likelihood of a step being compromised after a certain time, or a probability. The current formulation is kind of like specifying the average of a geometric distribution, which would imply that all steps have a probability. 

> How would uncompromising work? Would specific steps be non-uncompromisable?
No steps can be 'uncompromised' currently. Previously, defense steps would make steps 'uncompromised' but we changed it after much discussion.  This is related to the question of whether certain steps should have persistence.  
Steps never going back to being 'false' ultimately makes the entire simulator logic monotonic which makes for simpler expressions overall. For more expressive logic you have to turn to non-monotonic logic formulations, which become more complicated (you have to choose a solution to the frame problem).

> what happens if the simulation enables compromised steps to be re-compromised, i.e. the agent has the option of redoing things. This is a requested feature for the simulator to support.

[...]

PR #270, opened 2025-12-23
Implement actions and effect in MAL simulator

mrkickling:
This PR implements actions and effects in the MAL Simulator.

It uses the new concept of actions and effects in MAL. (causal mode)

If an agent performs an action, and the action has effects as children, those effects will be performed.
It will continue do perform children of effects as long as they are 'traversable'.

This solves issue #86


sandorstormen:
Have the appropriate changes been made to the `mal-toolbox` to accommodate effect and action?

sandorstormen (approved):
Seems fine to me.

PR #272, opened 2026-01-06
feat: Order of Performed Steps

sandorstormen:
Adds a field to all agent states.
```python

performed_nodes_order: dict[int, frozenset[AttackGraphNode]]
```

Initially compromised nodes will be a set under the `0` key. All other compromised nodes will be under subsequent time keys.

### DIFF
--- a/malsim/mal_simulator/attacker_state.py
+++ b/malsim/mal_simulator/attacker_state.py
@@ -60,7 +60,9 @@ class MalSimAttackerState(MalSimAgentState):
             self, 'ttc_value_overrides', MappingProxyType(state['ttc_value_overrides'])
         )
         object.__setattr__(
-            self, 'performed_nodes_order', MappingProxyType(state['performed_nodes_order']),
+            self,
+            'performed_nodes_order',
+            MappingProxyType(state['performed_nodes_order']),
         )
         # set other frozen attributes
         for key, value in state.items():


malsim/mal_simulator/agent_state.py
```diff
@@ -215,12 +230,22 @@ def create_defender_state(
...
+        if len(step_enabled_defenses) > 0:
+            performed_nodes_order[0] = frozenset(step_enabled_defenses)
     else:
         previous_enabled_defenses = previous_state.performed_nodes
         previous_compromised_nodes = previous_state.compromised_nodes
         previous_observed_nodes = previous_state.observed_nodes
         action_surface_additions = frozenset()
         action_surface_removals = step_enabled_defenses
+        performed_nodes_order = previous_state.performed_nodes_order
+        if time is not None:
```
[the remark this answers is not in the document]
mrkickling: previous_state is already checked in the outer else-statement I think

malsim/mal_simulator/agent_state.py
```diff
@@ -130,6 +135,11 @@ def create_attacker_state(
         ttc_value_overrides = previous_state.ttc_value_overrides
         impossible_step_overrides = previous_state.impossible_step_overrides
         compromised_nodes = previous_state.performed_nodes | step_compromised_nodes
+        performed_nodes_order = previous_state.performed_nodes_order
+        if len(step_compromised_nodes) > 0:
+            performed_nodes_order[previous_state.iteration] = frozenset(
```
mrkickling: should it not be previous_state.iteration + 1? Just want to make sure its not a typo.

malsim/mal_simulator/agent_state.py
```diff
@@ -24,6 +24,8 @@ class MalSimAgentState:
     action_surface: frozenset[AttackGraphNode]
     # Contains all nodes that this agent has performed successfully
     performed_nodes: frozenset[AttackGraphNode]
+    # Contains the order of performed nodes
+    performed_nodes_order: dict[int, frozenset[AttackGraphNode]]
```
mrkickling: performed_nodes_order in your implementation is not immutable like all other datastructures in the AgentState. Should be MappingProxyType instead.

malsim/mal_simulator/agent_state.py
sandorstormen: The results agree with the `MalSimulator.recording`.

sandorstormen: Please mark this comment as resolved if you are satisfied with my answer @mrkickling.

mrkickling:
re_based_

PR #275, opened 2026-01-07
Split simulator class

kasanari:
The MalSimulator class is a huge mess of state and functions, and has a circular dependency with the agent_states module.

This draft PR is an attempt to try to untangle the class somewhat.


mrkickling:
Yes, I feel that this is a bit overwhelming. Better to split up one thing at a time when possible.

kasanari:
I made an issue out of the circular dependency, since I think it is a big blocker for improving the structure:
https://github.com/mal-lang/mal-simulator/issues/276

PR #279, opened 2026-01-08
Break Up Simulator Class

kasanari:
Same thing as the other PR but this goes a step further and breaks up the sim parameter into the fields. If the reaction to any of these functions is "boy that seems like a lot of parameters", I agree with that reaction, but I will also note that this is was already the case before I made the change. The dependencies were just masked by the sim object. These functions can very likely be split into separate functions that are more focused.

@mrkickling there is one test failing that I am unsure of what is wrong. I would appreciate if you can take a look at it.


mrkickling:
Fix in #280

mrkickling:
I think just the split up will make it much more manageable when going through the code base.  So it is a good start.

kasanari:
Go for it :+1:

PR #282, opened 2026-01-14
Split up into modules

mrkickling:
- Create the MalSimulatorState that stores state in the simulator
- Split functions into more modules
- Make sure each functions gets what it minimally needs
- Remove circular dependencies through changes of responsibility in modules

Huge PR, focus on the file names and the simulator.py. Can anything more be stripped or moved from there? Should the split up look different? What modules should be there and what names should they have?

@kasanari perhaps I did some double work from what you did

### DIFF
--- a/malsim/mal_simulator/simulator.py
+++ b/malsim/mal_simulator/simulator.py
@@ -47,9 +47,11 @@ from malsim.mal_simulator.types import (
 from malsim.scenario.scenario import Scenario
 from malsim.mal_simulator.attacker_state_factories import create_attacker_state
 from malsim.mal_simulator.defender_state_factories import create_defender_state
-from malsim.mal_simulator.simulator_state import MalSimulatorState
+from malsim.mal_simulator.simulator_state import (
+    MalSimulatorState,
+    create_simulator_state,
+)
 from malsim.config.sim_settings import MalSimulatorSettings, RewardMode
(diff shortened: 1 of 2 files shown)


mrkickling:
One failing test, can not figure out why atm.

malsim/mal_simulator/agent_state.py
```diff
@@ -1,16 +1,13 @@
 from __future__ import annotations
```
kasanari: ditto on splitting attacker/defender

malsim/mal_simulator/reset_agent.py
```diff
@@ -0,0 +1,78 @@
...
+
+    # Revive all agents and reset reward
+    alive_agents = set(agent_settings.keys())
+    agent_states: AgentStates = {}
+    agent_rewards: AgentRewards = {}
+    pre_compromised_nodes: set[AttackGraphNode] = set()
+
+    # Create new attacker agent states
+    for attacker in agent_settings.values():
+        if isinstance(attacker, AttackerSettings):
```
kasanari: A lot of these checks could be avoided if agent settings was a typed structure, or did the filtering internally, e.g. agent_settings.defender_settings. Alternatively, for attacker in filter(lambda x: isinstance(x, AttackerSettings), agent_settings.values())

mrkickling: I'll start with a filter so we keep agent settings be a dict

PR #287, opened 2026-01-22
286 let users provide agents in malsimulator init

mrkickling:


malsim/mal_simulator/simulator.py
```diff
@@ -172,6 +174,24 @@ def __init__(
...
+        self._agent_settings: dict[str, AttackerSettings | DefenderSettings] = (
+            agent_settings or {}
+        )
+        self._agent_states: AgentStates = {}
+        self._alive_agents: set[str] = set()
+        self._agent_rewards: AgentRewards = {}
+
+        if self._agent_settings:
+            # register agents if they were given
+            self._agent_states, self._alive_agents, self._agent_rewards = reset_agents(
```
kasanari: This is essentially doing the same thing as `reset()`, right? If so, why not call it instead?

mrkickling: Maybe I should just add a helper `register_agents` and use that.

kasanari: I made a PR to this branch that does basically that.

PR #300, opened 2026-01-30
feat: Add EventLogger which makes use of language defined Detectors

### LINKED ISSUE #174: Incorporate Nikolaos detectors
Could be related to #173 .

mrkickling:
Detectors in MalSimulator which are used to generate event logs which can be observed by defenders

### DIFF
--- a/malsim/mal_simulator/defender_state.py
+++ b/malsim/mal_simulator/defender_state.py
@@ -1,9 +1,10 @@
 from __future__ import annotations
 from collections.abc import Set
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import Any
 from maltoolbox.attackgraph import AttackGraphNode
 from malsim.config.agent_settings import DefenderSettings
+from malsim.mal_simulator.event_logger import LogEntry
 from malsim.mal_simulator.agent_state import AgentState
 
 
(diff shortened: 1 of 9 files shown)


malsim/mal_simulator/defender_state.py
```diff
@@ -22,6 +23,7 @@ class MalSimDefenderState(MalSimAgentState):
     step_observed_nodes: frozenset[AttackGraphNode]
 
     # Agent specific rules for node properties
+    logs: EventLogger = EventLogger()
```
kasanari: I am a bit wary of this class (which carries it's own internal state and history) is included in the DefenderState. I think I see what is being desired, but the schema should be defined.

malsim/mal_simulator/event_logger.py
```diff
@@ -0,0 +1,133 @@
...
+    """A single log entry produced by a detector."""
+
+    timestep: int
+    detector_name: str
+    asset_name: str
+    attack_step_name: str
+    context_assets: dict[str, str]
+
+
+class EventLogger:
```
kasanari: The function is this class needs to be better defined. It is as far as I understand it a buffer for previous observations, but such temporal information should not be specific to just this data object.

PR #309, opened 2026-02-20
feat: Add wait prob to random agent

sandorstormen:
PR according to @mrkickling's [comment](https://github.com/mal-lang/mal-simulator/pull/308#issuecomment-3928205057).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -179,3 +179,5 @@ uv.lock
 
 # Profiling
 prof/*
+
+*.sh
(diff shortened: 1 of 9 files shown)


mrkickling:
Looks good, failing tests though

sandorstormen:
> Looks good, failing tests though

IMO, we should not have test for RandomAgent that check a specific seed. It's not checking that the RandomAgent is actually random.

PR #315, opened 2026-02-26
Settings revision

kasanari:
What was meant to be a small revision of the way settings are handled turned into quite a large amount of changes... 

The main change here is that settings that used to be global (rewards, false positives, etc.) are now stored in sim_settings. However, I think it might be more proper that these should be stored as a separate agent entry that overrides those with empty settings or something.

@mrkickling one test is failing because the attack steps do not have assets in them. Could you take a look as to why this is?


kasanari:
Now all the tests and checks pass

malsim/mal_simulator/simulator.py
```diff
@@ -129,21 +126,8 @@ class MalSimulator:
     def __init__(
         self,
         attack_graph: AttackGraph,
+        agents: Iterable[AttackerSettings[AttackGraphNode] | DefenderSettings],
```
mrkickling: AttackerSettings has T = AttackGraphNode. this is the type of the entrypoints/goals I think. It should also be possible to give full names, not just AttackGraphNodes. Should the correct type be:

`agents: Iterable[AttackerSettings[AttackGraphNode | str] | DefenderSettings],`

or what do you think?

malsim/mal_simulator/run_simulation.py
```diff
@@ -11,6 +11,7 @@ def run_simulation(sim: MalSimulator) -> dict[str, list[AttackGraphNode]]:
 
     Return selected actions by each agent in each step
     """
+    agents = sim.agent_settings
```
mrkickling, line 13: the `agents` variable should be used instead of `sim.agent_settings` below this line in this function, or removed.

malsim/mal_simulator/rewards.py
```diff
@@ -3,72 +3,90 @@
...
+from malsim.config.agent_settings import Attacker, Defender
 from malsim.mal_simulator.defender_state import MalSimDefenderState
 from malsim.mal_simulator.attacker_state import MalSimAttackerState
 from malsim.mal_simulator.graph_utils import node_reward
 from malsim.config.sim_settings import RewardMode, TTCMode
 from malsim.mal_simulator.ttc_utils import TTCDist
 
 
-def defender_step_reward(
+def defender_step_reward_fn(
```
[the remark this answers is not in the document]
mrkickling: Can we then at least have a docstring for defender_step_reward_fn which explains that it returns different reward functions depending on settings. or whichever explanation is correct?

malsim/scenario/scenario.py
```diff
@@ -37,6 +38,7 @@
     'attacker_agent_class',
     'defender_agent_class',
     'attacker_entry_points',
+    'rewards',
```
mrkickling, line 41: we should add:
    'observable_steps',
    'actionable_steps',
    'false_positive_rates',
    'false_negative_rates',
to this list so the user gets a more meningful log when it crashes

malsim/scenario/scenario.py
```diff
@@ -142,29 +141,53 @@ def from_dict(cls, scenario_dict: dict[str, Any]) -> Scenario:
...
+        sim_settings = scenario_dict.get('sim_settings', {})
+        sim_settings = (
+            asdict(MalSimulatorSettings(**sim_settings))
+            if not isinstance(sim_settings, MalSimulatorSettings)
+            else asdict(sim_settings)
+        )
+
+        # for legacy reasons, also support rewards, false_positive_rates,
+        # and false_negative_rates at the top level of scenario dict
+        if 'rewards' in scenario_dict:
```
mrkickling: Ah ok, I see here that they are deprecated in a different way. Then I am fine with it.

malsim/scenario/scenario.py
```diff
@@ -142,29 +146,57 @@ def from_dict(cls, scenario_dict: dict[str, Any]) -> Scenario:
...
+            logger.warning(
+                "Found 'false_negative_rates' in scenario dict. "
+                "This should be moved under 'agent_settings'"
+            )
+            del scenario_dict['false_negative_rates']
+
         return Scenario(
             lang_file=scenario_dict['lang_file'],
-            agent_settings=agent_settings,
+            agents=agent_settings,
```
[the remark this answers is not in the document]
mrkickling: Im ok with calling it agents

mrkickling:
Take a look at the comments and then when they are resolved or explained in some way I will approve.

malsim/mal_simulator/simulator.py
kasanari: AttackerSettings is already bounded to take either `AttackGraphNode | str`, so then `___init___` or the Simulator object itself would also be generic with bounds `AttackGraphNode | str`. The problem is that somewhere there has to be a check whether we are using names or objects, and ideally that check should happen very early. I think the main benefit of using `AttackGraphNode` (or perhaps a lighter type that is only name+asset_id) is that it assures that the ID is in the attack graph. However, this checking should probably happen at some external layer IMO, like how it is currently done in the Scenario class. I could do a check in `__init__` too though, to refine `AttackGraphNode | str` into just `AttackGraphNode`.

malsim/scenario/scenario.py
kasanari: i removed it since it was a bit hacky

malsim/mal_simulator/simulator.py
mrkickling: I am not sure, but I think it would be nice to be able to run this example: https://github.com/mal-lang/mal-simulator/blob/34deb8c7d36a814ce970f17454518e465d001810/examples/example_no_scenario.py with full names instead of having to do `attackgraph.get_node_by_full_name` for each node?

I imagine that some people will create scenarios while some will create MalSimulators without scenarios and both should be as easy as possible.

malsim/mal_simulator/rewards.py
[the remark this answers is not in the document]
kasanari: Done

tests/agents/test_path_finders.py
```diff
@@ -8,12 +8,10 @@
 def test_path_finding() -> None:
     scenario_file = 'tests/testdata/scenarios/traininglang_scenario.yml'
     scenario = Scenario.load_from_file(scenario_file)
-    sim = MalSimulator.from_scenario(scenario, register_agents=False)
```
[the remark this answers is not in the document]
kasanari: It is possible to set agents=()

malsim/mal_simulator/simulator.py
kasanari: I have added validation for the __init__ step now. However, one still has to supply the attack_graph to __init__, which means that the caller of the function could do the same validation. I generally prefer to push validation upwards rather than downwards, but we can keep it like this for now.

mrkickling (approved):

PR #326, opened 2026-03-12
remove Optional

kasanari:
Since the Optional type is deprecated since 2.10, and the mal-simulator is declared to not support <2.10, i think it would be good to replace it with the new style of declaring optional parameters.


mrkickling:
Had no clue, good.

mrkickling (approved):

PR #327, opened 2026-03-16
make extend recurse into mappings

kasanari:
Change behavior of extend to recurse into mappings, since the current extend is not super useful otherwise with the new settings revision.

This changes the behavior slightly, however, since before it would override the agents dict completely rather than extend it.


kasanari:
Added an allowance for setting the key to None, thus ignoring the extend

mrkickling (approved):

PR #338, opened 2026-03-18
feat: remove step_x fields from attacker and defender

mrkickling:
Question:
- I created 'properties' so everything doesn't break, but that kind of defeats the purpose of removing the fields and also it has some code duplication. Should we let users themselves run 'diffs' or give helpers?
- Should we remove the AgentState class?


sandorstormen:
> Should we let users themselves run 'diffs' or give helpers?

Give helpers!

malsim/mal_simulator/agent_state.py
```diff
@@ -21,13 +20,8 @@ class AgentState:
...
-    # Contains the nodes performed successfully in the last step
-    step_performed_nodes: Set[AttackGraphNode]
-    # Contains possible nodes that became available in the last step
-    step_action_surface_additions: Set[AttackGraphNode]
-    # Contains nodes that became unavailable in the last step
-    step_action_surface_removals: Set[AttackGraphNode]
-    # Contains nodes that became unviable in the last step by defender actions
-    step_unviable_nodes: Set[AttackGraphNode]
+
+    unviable_nodes: Set[AttackGraphNode]
```
kasanari, line 24: I assume this will be removed later with the viability removal PR?

kasanari (approved):

malsim/mal_simulator/agent_state.py
mrkickling: Yes most likely.

PR #339, opened 2026-03-20
feat: don't propagate viability to compromised nodes

mrkickling:


malsim/mal_simulator/graph_state.py
```diff
@@ -56,7 +56,7 @@ def compute_initial_graph_state(
         impossible_attack_steps = get_impossible_attack_steps(graph.attack_steps, rng)
 
     viability_per_node = calculate_viability(
-        graph, enabled_defenses, impossible_attack_steps
+        graph, enabled_defenses, impossible_attack_steps, set()
```
kasanari, line 59: Should this not be based on the entrypoint set?

malsim/mal_simulator/graph_processing.py
```diff
@@ -137,10 +138,10 @@ def _propagate_viability_from_node(
         is_viable = evaluate_viability(
             child, viability_per_node, frozenset(), impossible_attack_steps
         )
-        if is_viable != viability_per_node[child]:
+        if is_viable != viability_per_node[child] and child not in compromised_nodes:
             viability_per_node[child] = is_viable
```
kasanari, line 142: Would it be possible to also remove this mutable reference while we are changing this function?

malsim/mal_simulator/graph_state.py
mrkickling: Good question, should it? If there is an entry point that is blocked by a pre-enabled defense, should the entry point 'children' be blocked? I guess it should according to the logic we want to implement.

malsim/mal_simulator/graph_processing.py
mrkickling: Do you mean we should not change the value of `viability_per_node` in there?

malsim/mal_simulator/graph_state.py
kasanari: Makes sense to me.

malsim/mal_simulator/graph_processing.py
kasanari: Yes. Such that `viability_per_node` can be declared as `Mapping[AttackGraphNode, bool]`.

mrkickling:
Replaced by #341

PR #341, opened 2026-04-13
feat: remove viability

mrkickling:
- Store enabled defenses in sim_state instead of calculating viability each step
- Find out if a node is blocked before adding it to action surface or compromising it
- Defense surface only contains non-enabled defenses (same as before but more explicit now)
- Persistence on attack steps
- Persistence on entry points even if they are not performed at start
- Keep old viability calculation functions (this explains the lack of deleted lines)

Note:
- A lot of the test_example_scenario tests changed and they are hard to understand. I have decided to trust that their change was just due to the added persistence, after writing some more attack surface tests to make sure it is calculated correctly.

### DIFF
--- a/malsim/mal_simulator/attack_surface.py
+++ b/malsim/mal_simulator/attack_surface.py
@@ -64,14 +64,10 @@ def get_attack_surface(
     from_nodes = from_nodes if from_nodes is not None else performed_nodes
 
     skip_compromised = settings.skip_compromised
-    skip_unviable = settings.skip_unviable
     skip_unnecessary = settings.skip_unnecessary
 
     def uncompromised(node: AttackGraphNode) -> bool:
         return not (skip_compromised and node in performed_nodes)
-
-    def viable(node: AttackGraphNode) -> bool:
-        return not node_is_blocked(sim_state, node)
 
     def necessary(node: AttackGraphNode) -> bool:
         return node_is_necessary(sim_state, node)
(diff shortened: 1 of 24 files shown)


malsim/mal_simulator/attack_surface.py
```diff
@@ -87,8 +83,8 @@ def in_attack_surface(node: AttackGraphNode) -> bool:
         is_action = node.causal_mode != 'effect'
         return (
             is_action
+            and not node_is_blocked(sim_state, node)
```
kasanari, line 86: Seems simpler to have the function be called `node_is_actionable`, to not have to negate it, but that name is taken already...

kasanari (approved):

sandorstormen:
Are there checks to make sure the graphical gym env stays the same with this PR?

malsim/mal_simulator/attack_surface.py
mrkickling: Yes, I think a few of the 'concepts' can be merged at some points as well. Actionability could be defined as not blocked by parent and not blocked by actionability filter or something..

mrkickling:
> Are there checks to make sure the graphical gym env stays the same with this PR?

I haven't written any additional tests for the graphical gym env specifically

sandorstormen:
> I haven't written any additional tests for the graphical gym env specifically

It would be good if you looked it over to see if anything has changed, which shouldn't change.

mrkickling:
Sure, I am not sure exactly what I should test in the graphical gym env though. You had written some tests for it before.

Since the gym env uses the MalSimulator and the MalSimulator will work the same (except for persistant steps/entry points) I don't expect anything to change in the graphical gym env specifically.

PR #345, opened 2026-05-07
Change default reward mode of attacker

sandorstormen:


kasanari (approved):

kasanari:
I think having `reward_mode` be `RewardMode | None` with None as the default (i.e. no rewards are calculated unless the user explicitly specifies that they want some) might be a better option, but that would be a bigger change.

sandorstormen:
> I think having `reward_mode` be `RewardMode | None` with None as the default (i.e. no rewards are calculated unless the user explicitly specifies that they want some) might be a better option, but that would be a bigger change.

I mean anything other than cumulative reward mode is better for the attacker, in my mind. Where did we land on stickiness? I think this affects whether the attacker could potentially care about maintain his/hers level of compromise in the system.

mrkickling (approved):

PR #350, opened 2026-05-11
Multiple potential entrypoint sets

mrkickling:
Add the option to give more than one set of entry points in the scenario and the scenario file.

In scenario file previously:
```
  'Attacker1':
    type: attacker
    entry_points:
      - 'entrypoint1'
      - 'entrypoint2'
```

This is still supported, but additionally you can give multiple sets/lists:

```
  'Attacker1':
    type: attacker
    entry_points:
      - ['entrypoint1', 'entrypoint2']
      - ['entrypoint3', 'entrypoint4']
```

When running a simulation with this scenario, the attacker agent will be given one of the specified entry point lists, so either ['entrypoint1', 'entrypoint2'] or ['entrypoint3', 'entrypoint4'].
The rng is the same one seeded by mal sim settings so determinism can be fulfilled.

This also works in the AttackerSettings where you can now either give a set of entry point nodes or a tuple containing multiple sets of entry point nodes.

### DIFF
--- a/tests/test_mal_simulator.py
+++ b/tests/test_mal_simulator.py
@@ -1771,3 +1771,25 @@ def test_simulator_multiple_entry_point_sets_in_attacker_settings(model: Model) 
         sim.get_node('OS App:fullAccess'),
         sim.get_node('Program 2:fullAccess'),
     }
+
+
+def test_simulator_multiple_entry_point_sets_scenario():
+    """
+    Test that if a scenario has an attacker with multiple entry point sets,
+    one set is picked when initializing the simulator
+    """
+    scenario = Scenario.load_from_file(
+        'tests/testdata/scenarios/bfs_vs_bfs_scenario_multiple_entrypoint_sets.yml'
+    )
+
+    chosen_entry_points = list()
+    for i in range(10):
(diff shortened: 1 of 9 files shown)


tests/test_mal_simulator.py
```diff
@@ -1740,3 +1740,66 @@ def test_simulator_sim_settings_from_scenario(
...
+    for i in range(10):
+        sim = MalSimulator.from_scenario(
+            scenario, sim_settings=MalSimulatorSettings(seed=i)
+        )
+        attacker_state = sim.agent_states['attacker1']
+        assert isinstance(attacker_state, AttackerState)
+        chosen_entry_points.add(frozenset(attacker_state.entry_points))
+
+    # Make sure not all entrypoints are the same across all runs
+    assert len(chosen_entry_points) == 2, (
```
sandorstormen, line 1795: Fantastic test! It actually tests the functionality of the feature.

sandorstormen (approved):
LGTM

tests/test_mal_simulator.py
mrkickling: As always Sandor!

PR #352, opened 2026-05-12
Heuristic Agents with Gym Env

sandorstormen:
Allows for using heuristic agents with graphical gym env.

### DIFF
--- a/malsim/envs/graph/README.md
+++ b/malsim/envs/graph/README.md
@@ -6,6 +6,10 @@ The Graphical Gym Interface provides a graph-structured observation and action s
 
 The interface represents the attack graph as a structured observation with nodes (assets, steps, associations, logic gates) and edges (relationships between nodes). This graph structure is encoded using Coordinate (COO) format for efficient representation and processing.
 
+### Scenario settings
+
+**Important**: To be able to use any of the wrappers here with a [scenario](https://github.com/mal-lang/mal-simulator/wiki/Scenarios), you need to set `agent_class: null` in the YAML file. Or if creating it programatically, the policy/agent's class need to be `None`. Otherwise it will be assumed to be a heuristic agent.
+
 ## Core Concepts
 
 ### Observation Structure (`MALObsInstance`)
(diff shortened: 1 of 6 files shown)


malsim/envs/graph/graph_env.py
```diff
@@ -234,6 +236,17 @@ def __init__(
         self.possible_agents = list(self.sim.agent_states)
         self.agents = list(self.sim.alive_agents)
 
+        self.non_gym_agents: AgentSettings = {
```
kasanari: I would maybe suggest `internal` and `external` agents, rather than tying the function to the interface.

kasanari (approved):

malsim/envs/graph/README.md
```diff
@@ -6,6 +6,10 @@ The Graphical Gym Interface provides a graph-structured observation and action s
 
 The interface represents the attack graph as a structured observation with nodes (assets, steps, associations, logic gates) and edges (relationships between nodes). This graph structure is encoded using Coordinate (COO) format for efficient representation and processing.
 
+### Scenario settings
+
+**Important**: To be able to use any of the wrappers here with a [scenario](https://github.com/mal-lang/mal-simulator/wiki/Scenarios), you need to set `agent_class: null` in the YAML file. Or if creating it programatically, the policy/agent's class need to be `None`. Otherwise it will be assumed to be a heuristic agent.
```
kasanari, line 11: why not define an agent class like `learner` or `gym` or `external`? I forget if there is an enum somewhere with these values.

sandorstormen: Beacuse I will then have to create a `DecisionAgent` class that functions the same as `PassiveAgent`, but has a different meaning. There's no enum of agent classes, but there is a mapping between string and the actual class that selects the next action. It would be awkward to implement a class, whose functionality is meaningless, but carries meaning in its exsistance.

malsim/envs/graph/graph_env.py
sandorstormen: Good suggestion!

mrkickling (approved):

PR #363, opened 2026-08-04
Implement Simulator with Model Effects

sandorstormen:
This PR implements the `DynaMalSimulator` which enacts the model effects that attack steps have. This is intended to implement DynaMAL-like functionality, The `DynaMalSimulator` uses the `MalSimulator` as its parent class and changes the way the simulator resets and steps.

## Differences from _DynaMAL_

_DynaMAL_ respects the order in which the reaching clauses are written, as well as the sentences within the reaching clauses. In the DynaMalSimulator we don't respect order and instead we execute the modifications to the instance model as effects of having preformed a step. The order is: Additive effects, then subtractive effects and lastly step effects.

## State of the DynaMalSimulator

The design choice has been that purity of functions outside of `dyna_mal_simulator/simulator.py` is broken. This is because dynamic targets may reference assets/associations that are newly created. So model effects have to be parsed **after** any previous model effects have been applied. This preserves the left-to-right order of dynamic reaching clauses in _DynaMAL_

## Partial Attack Graph Regeneration
Partial attack graph regeneration is used to stay compatible with the functionality implemented for the `MalSimulator`. This is because large parts of the code base assumes that the attack graph is not re-generated and that the same `AttackGraphNode` objects are used for subsequent simulations. Minor improvements to efficiency compared to complete attack graph regeneration is just bonus.

## Interpretation of *DynaMAL*-like Grammar
Consider this dynamicExampleLang:
```
category Examples {
  asset A {
    | add
      A> <base> / <anchor set>.<termination>
    | remove
      R> <base> / <anchor set>.<termination>
    | connect
      A> <base> / ~<anchor set>.<termination>
    | disconnect
      R> <base> / ~<anchor set>.<termination>
  }
}
```
The following is the interpretation that is implemented

Collect the set of assets `<anchor set>`, based on the assets which can be reached from the `<base>`. For each asset in the `<anchor set>`, add an asset of type given by `<termination>`. Then associate the asset to the newly created asset through the field `<termination>`.

Collect the set of assets `<anchor set>`, based on the assets which can be reached from the `<base>`. For each asset in the `<anchor set>`, remove all the assets associated by `<termination>`. 

Collect the set of assets `<base>`, based on the assets which can be reached from the `A` asset. Then, collect the set of assets `<anchor set>` in the same manner. For each asset in the `<anchor set>`, associate it to the assets in `<base>` via `<termination>`.

Collect the set of assets `<anchor set>`, based on the assets which can be reached from the `<base>`. For each asset in the `<anchor set>`, remove association to assets associated by the field `<termination>`.

## Dynamic Errors
Whenever a model effect is applied where non-existent asset/association is removed, the error is logged but the simulation continues. When adding an asset/association, where it's disallowed by the constraints set out by the association in the language, the error is logged and the simulation continues. Whenever a model effect references a non-existent association by field name, the error is logged and the simulation continues. 

## Model Effect Application Algorithm
Model effects are applied by:
1. Modify the instance model for all _additive_ modifications.
2. Modify the instance model for all _subtractive_ modifications.
3. Partial generation of the attack graph is done based on modifications.
4. The graph state is updated based on any newly added attack graph nodes.
  a. TTC values are sampled for **new** attack steps (depending on TTC mode).
  b. Status of new defense steps are sampled (depending on `run_defense_step_bernoullis`).
  c. **New** attack steps are sampled to decide if they are impossible to compromise (depending on `run_attack_step_bernoullis`).
  d. Necessity is re-calculated for **all** attack graph nodes. (Potentially inefficient)

## Dynamic Step
Each agent triggers the model effects when performing a step.

### Dynamic Defender Step
Model effects are applied directly after enabling a defense step.

### Dynamic Attack step
The step check looks like as follows:
1. Check if node is entry point or traversable
2. If so, attempt compromise taking into account the current TTC mode.
3. If attacker is able to compromise: 
  a. **Apply model effects**.  <- Added by this PR
  b. Handle `@effect`-steps

## Dynamic Reset

The reset function in the `DynaMalSimulator` class undoes modifications to the instance model before doing "regular reset". The steps are:

1. Apply model effects in reverse. Additions become subtractions, and vice versa.
2. Partial generation of the attack graph is done based on the reversed modifications.
3. Continue with "regular reset".

#### [Parent-PR in the `tree-sitter-mal` repo](https://github.com/mal-lang/tree-sitter-mal/pull/43)
Defines the grammar with _DynaMAL_-like syntax.

#### [Parent-PR in the `mal-toolbox`](https://github.com/mal-lang/mal-toolbox)
Compiles and parses the grammar in the `tree-sitter-mal` PR.

### DIFF
--- a/malsim/dyna_mal_simulator/__init__.py
+++ b/malsim/dyna_mal_simulator/__init__.py
@@ -1,0 +1,5 @@
+"""DynaMALSimulator related modules."""
+
+from .simulator import DynaMalSimulator
+
+__all__ = ['DynaMalSimulator']
--- a/malsim/dyna_mal_simulator/attacker_step.py
+++ b/malsim/dyna_mal_simulator/attacker_step.py
@@ -1,0 +1,135 @@
+from __future__ import annotations
+from typing import TYPE_CHECKING
+import logging
+
+import numpy as np
+
+from maltoolbox.attackgraph import AttackGraphNode
+
+from malsim.config.sim_settings import TTCMode
+from malsim.dyna_mal_simulator.model_effects import execute_model_effects
+from malsim.dyna_mal_simulator.simulator_state import DynaMalSimulatorState
+from malsim.mal_simulator.graph_utils import (
+    node_is_traversable,
+)
+from malsim.mal_simulator.attacker_step import (
+    attacker_step_effects,
+)
+from malsim.mal_simulator.state_query import node_ttc_value
+from malsim.mal_simulator.ttc_utils import TTCDist
+
+if TYPE_CHECKING:
+    from malsim.mal_simulator.attacker_state import AttackerState
+
+logger = logging.getLogger(__name__)
+
+
+def dyna_attacker_step(
+    sim_state: DynaMalSimulatorState,
+    agent: AttackerState,
+    nodes: list[AttackGraphNode],
+    rng: np.random.Generator,
+) -> tuple[list[AttackGraphNode], list[AttackGraphNode], DynaMalSimulatorState]:
+    """Compromise attack step nodes with attacker
+
+    Args:
+    agent - the agent to compromise nodes with
+    nodes - the nodes to compromise
+
+    Returns: two lists with compromised, attempted nodes
+    """
+
+    logger.debug('Stepping with agent %s', agent.name)
+    successful_compromises: list[AttackGraphNode] = []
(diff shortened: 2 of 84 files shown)


sandorstormen:
Requesting review from everybody for visibility and to allow comments/suggestions to be made before I merge the PR.

